PR #2, opened 2016-02-09
Using a local database to store model details.

slimjim777:
Adding database connection for storing model information.
Switching to using an environment struct to pass config and database information.
Tool to create the 'model' database table.
Implement the models API call to get data from the database.
Additional unit tests to check the error conditions.


arapulido:
There is a missing dependency that made the Travis CI tests fail. 

Also, do you know why my commits are in the pull request if they are already merged?

slimjim777:
Yes, I saw the issue with the dependencies. I'll update that tomorrow. 
I think the travis changes are there because I merged HEAD into my branch and later did a rebase. Those files haven't changed. I had been working on the master branch, so maybe that confused things as well. 
Hopefully things should be more tidier after this pull request goes through.
Thanks 
James






On Tue, Feb 9, 2016 at 10:47 AM -0800, "Ara Pulido" <notifications@github.com> wrote:












There is a missing dependency that made the Travis CI tests fail. 



Also, do you know why my commits are in the pull request if they are already merged?

service/handlers.go
```diff
@@ -83,3 +124,27 @@ func SignHandler(w http.ResponseWriter, r *http.Request) {
...
+func ModelsHandler(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+	w.WriteHeader(http.StatusOK)
+
+	var models []ModelDisplay
+
+	dbModels, err := Environ.DB.ListModels()
+	if err != nil {
+		errorMessage := fmt.Sprintf("Error fetching the models: %v", err)
+		formatModelsResponse(false, errorMessage, nil, w)
```
arapulido: Shall we return a HTTP 5xx code instead in this case?

service/database.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// OpenSysDatabase Return an open database connection
+func OpenSysDatabase(driver, dataSource string) *DB {
+	// Open the database connection
+	db, err := sql.Open(driver, dataSource)
+	if err != nil {
+		log.Fatal(err)
+	} else {
+		log.Println("Database opened successfully.")
```
arapulido: sql.Open does not actually opens the database, so this will succeed even if there is no database available, and then it will fail elsewhere when trying to access it (when accessing the models endpoint)

"Perhaps counter-intuitively, sql.Open() does not establish any connections to the database, nor does it validate driver connection parameters. Instead, it simply prepares the database abstraction for later use. The first actual connection to the underlying datastore will be established lazily, when it’s needed for the first time. If you want to check right away that the database is available and accessible (for example, check that you can establish a network connection and log in), use db.Ping() to do that, and remember to check for errors:

err = db.Ping()
if err != nil {
	// do something here
}
"
(As explained at: http://go-database-sql.org/accessing.html)

service/modelmanager.go
```diff
@@ -0,0 +1,71 @@
...
+	return err
+}
+
+// ListModels fetches the full catalogue of models from the database.
+func (db *DB) ListModels() ([]Model, error) {
+	var models []Model
+
+	rows, err := db.Query(listModelsSQL)
+	if err != nil {
+		log.Fatal(err)
```
arapulido: This shouldn't be a Fatal error, as this is called from a HTTP handler, so if there is an error, the server will shutdown. Instead of that, it should return an error, so the handler can return an error to the user gracefully

PR #4, opened 2016-02-15
Added the basic web application to load the HTML template.

slimjim777:
Style the HTML using Ubuntu Vanilla Theme.
Use gulp and npm for building React-based web UI.


README.md
```diff
@@ -1,3 +1,4 @@
+[![Build Status][travis-image]][travis-url] [![Coverage Status][coveralls-image]][coveralls-url]
```
arapulido: We still don't have a coveralls account/set up. I don't think this would be harmful, but not needed at this point

arapulido:
+1

PR #5, opened 2016-02-17
Fix empty data checks for the SignHandler.

arapulido:
As you can see in the [documentation for http Request](https://golang.org/pkg/net/http/#Request), for server requests the Request Body is always non-nil, but will return EOF immediately when no body is present.

This commit checks for io.EOF **after** the response has been decoded (this will return io.EOF if the body was empty).

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -22,6 +22,7 @@ package service
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"log"
 	"net/http"
 )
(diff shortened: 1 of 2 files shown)


slimjim777:
Testing for a nil Body doesn't seem necessary then, since the body is always non-nil. I'll update the check on the Body as suggested.

arapulido:
This one is closed, but the other one is ready for merging: https://github.com/ubuntu-core/identity-vault/pull/6

Please, merge that one instead

arapulido:
And the nil body is necessary, for completeness (as someone could call that library method using nil)

PR #7, opened 2016-02-18
Validate the model against the accepted models before signing

slimjim777:
Check the model with the local data store and retrieve the location of the signing-key from the database.
Additional tests with mocks to simulate database errors.


service/handlers.go
```diff
@@ -107,6 +107,14 @@ func SignHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Validate the model by checking that it exists on the database
+	model, err := Environ.DB.FindModel(assertions.Brand, assertions.Model, assertions.Revision)
+	if err != nil {
+		errorMessage := "Cannot find model with the matching brand, model and revision."
+		formatSignResponse(false, errorMessage, "", w)
```
arapulido: We are not returning a HTTP code in this case.

arapulido:
+1, once we fix the small comment above.

PR #8, opened 2016-02-19
Table driven tests

arapulido:
Changed the tests for handlers to a table driven schema to avoid duplication.

Go recommends using table driven tests when possible: https://github.com/golang/go/wiki/TableDrivenTests

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -138,7 +138,7 @@ func SignHandler(w http.ResponseWriter, r *http.Request) {
 	signedText, err := ClearSign(dataToSign, string(privateKey), "")
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


slimjim777:
I understand the principle of table driven tests, and how they remove some code repetition. However, in these cases, I think that the tests have removed clarity and code readability. The example table-driven test is a good case as there are just two simple string variables (in and out), but in these cases we have a number of variables and they are not that straightforward.

For now, I think we'd get more clarity and code reduction be putting repeated code into a function that is called by the tests. However, since I'm still in the process of adding more tests, so I think it would be better to review this when we have all the functionality in place. We can then assess what the best approach for the tests should be.

arapulido:
Fair enough. The branch contained some improvements to the tests, though, that I would like to introduce. I will close this one and create a new one with those improvements.

PR #9, opened 2016-02-24
Add create, update and get methods for a model.

slimjim777:
Fix model retrieval from database.
Add update and get methods for a model.
Refactor the router to a separate function so it can be used in tests.
Add creation of new model to the API and UI.
Add filesystem-based keystore for the signing-keys.
Refactor of the model tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 .coverage
 *.log
 node_modules
+keystore/model*
+tags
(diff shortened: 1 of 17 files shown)


arapulido:
Looks good (although haven't checked Javascript bits)

It would be good in future merge proposals to split the content into several commits, as it would make the review easier (each of the lines in your commit message could have been a separate commit)

PR #10, opened 2016-03-03
API and UI to view, create and remove authorized public keys.

slimjim777:
Added AuthorizedKeys API method to list the available ssh keys.
Added creating and deleting of authorized ssh keys in the API.
Added UI for viewing, creating and deleting a public key.


server.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/ubuntu-core/identity-vault/service"
 )
 
+const sshKeysPath = "/.ssh/authorized_keys"
```
arapulido: Shouldn't this folder be configurable through the config file?

service/authkeys.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+// AuthorizedKeystore interface to manage the authorized ssh keys
+type AuthorizedKeystore interface {
+	List() []string
+	Add(string) error
+	Delete(string) error
+}
+
+// AuthorizedKeys interface to manage the authorized ssh keys
```
arapulido: Wrong comment?

server.go
```diff
@@ -35,6 +37,9 @@ func main() {
 		log.Fatalf("Error parsing the config file: %v", err)
 	}
 
+	// Initialize the authorized keys manager
+	env.AuthorizedKeys, err = service.InitializeAuthorizedKeys(sshKeysPath)
```
arapulido: If there is an error, we should manage it.

service/authkeys.go
```diff
@@ -0,0 +1,162 @@
...
+	var keys []string
+
+	// Turn on the read lock
+	auth.mu.RLock()
+	defer auth.mu.RUnlock()
+
+	// Read the authorized_keys file
+	file, err := os.Open(auth.path)
+	if err != nil {
+		return keys
```
arapulido: This should log a proper error. (i.e. authorized_keys file not found)

service/authkeys.go
```diff
@@ -0,0 +1,162 @@
...
+			break
+		}
+	}
+	return authKeyPosition, keys
+}
+
+// Add saves a new authorized ssh key
+func (auth *AuthorizedKeys) Add(key string) error {
+	// Check the public key
+	key = strings.Trim(key, " ")
```
arapulido: Maybe the trimming of the key also makes sense in the previous method?

service/handlers_keys.go
```diff
@@ -0,0 +1,131 @@
...
+
+// BooleanResponse is the JSON response from an API method, indicating success or failure.
+type BooleanResponse struct {
+	Success      bool   `json:"success"`
+	ErrorMessage string `json:"message"`
+}
+
+// AuthorizedKey is the JSON body with the public key.
+type AuthorizedKey struct {
+	PublicKey string `json:"device-key"`
```
arapulido: The JSON tag should be ssh-key or public-key. device-key is confusing with other term

service/handlers_keys.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		message := fmt.Sprintf("Error deleting a public key: %v", err)
+		log.Printf(message)
+		formatBooleanResponse(false, message, w)
+	} else {
+		formatBooleanResponse(true, "", w)
+	}
+}
+
+// AuthorizedKeyDeleteHandler removes a key from the authorized keys file.
```
arapulido: Wrong function header.

service/handlers_keys_test.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func TestAuthorizedKeyAddHandlerNilData(t *testing.T) {
+	// Use a new authorized_keys file
+	path := "../test/authorized_keys_new"
+	Environ = &Env{AuthorizedKeys: &AuthorizedKeys{path: path}}
+
+	result, _ := sendAddRequest(t, nil)
+	if result.Success {
+		t.Error("Expected an error, got success.")
```
arapulido: On this type of calls, where we get different errors on different places, shouldn't we check the error message expected as well?

arapulido:
Made some comments to the Go code.

server.go
slimjim777: That is the default location for the ssh files for Ubuntu systems (and most other OS's), so it didn't seem necessary to make it configurable.

service/authkeys.go
slimjim777: Not having an authorized_keys keys file isn't really an error. It just means that there are no keys authorized, so the empty list is returned.

server.go
arapulido: OK

service/authkeys.go
arapulido: ok

service/handlers_keys_test.go
slimjim777: Checking the message text for each error doesn't seem so robust and it means duplicating the error messages. I think that just leads to less manageable tests as we cannot change the error message text without changing the linked tests as well. This then gets even more messy if we need to introduce i18n into the Go messages.

I have some thoughts about returning message keys instead of actual messages, which will get converted into message text in the correct language in the UI. At that stage, I think we can revisit the tests.

arapulido:
Looks good, merging.

PR #11, opened 2016-03-08
Improved signing handler tests

arapulido:
This is based on the branch I created (and we closed) about table driven tests.
Instead of creating a table, I have extracted functionality for calling the signing handler into a function.

Apart of that, for all tests, I have included a check for the right error message when expected.

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -159,7 +159,7 @@ func SignHandler(w http.ResponseWriter, r *http.Request) {
 	signedText, err := ClearSign(dataToSign, string(privateKey), "")
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


slimjim777:
The main SignHandler refactoring looks good, but I'm not keen on checking the error message texts. This is going to introduce a lot of extra maintenance and duplication, especially if the error messages change.

Having started work on i18n part of the UI, it is very likely that the approach to error messages will change in the Go code so that we can deliver internationalized error messages to the UI. I agree that we need to work on refactoring the tests, but I think it is premature to introduce error message checking at this stage. I think it would be better to review this after the i18n changes land.

I've added a card for the refactoring of tests, so this doesn't get missed.

arapulido:
Sounds good. Closing then.

PR #12, opened 2016-03-11
i18n for the UI and API

slimjim777:


arapulido:
It looks good, but I have a question.

For replies that are not going to go through the UI (i.e. sign), shouldn't we add a message even if it is always in English?

For example:
formatSignResponse(false, "error-sign-empty", "", "", "", w)

In that case, when you request signing with an empty body, the JSON you get back doesn't have a person friendly message. In these cases I think we should add the errorMessage to the response.

Any reason why we don't?

arapulido:
For example, if I use curl to sign something (with an empty body) I would get:

ara@sushirider:~$ curl -X POST http://10.55.60.91:8080/1.0/sign
{"success":false,"error_code":"error-sign-empty","error_subcode":"","message":"","identity":""}

which is not very helpful.

PR #13, opened 2016-04-12
Use Ubuntu-core assertions module for core operations

slimjim777:


service/handlers.go
```diff
@@ -111,67 +104,66 @@ func SignHandler(w http.ResponseWriter, r *http.Request) {
...
+	// Check that we have a device-serial assertion (the details will have been validated by Decode call)
+	if assertion.Header("type") != "device-serial" {
+		w.WriteHeader(http.StatusBadRequest)
+		formatSignResponse(false, "error-decode-assertion", "error-invalid-type", "The assertion type must be 'device-serial'", nil, w)
 		return
 	}
 
-	// Format the assertions string
-	dataToSign, err := formatAssertion(assertions)
+	revision, err := strconv.Atoi(assertion.Header("revision"))
```
arapulido: Checking revision is not needed, it is already done in asserts.Assemble (called at Decode). If Decode has been successful we can assume a valid revision, and it can be retrieved using the interface function "Revision() int"

service/handlers.go
```diff
@@ -111,67 +104,66 @@ func SignHandler(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
-	// Validate the model by checking that it exists on the database
-	model, err := Environ.DB.FindModel(assertions.Brand, assertions.Model, assertions.Revision)
-	if err != nil {
-		w.WriteHeader(http.StatusNotFound)
-		formatSignResponse(false, "error-model-not-found", "", "Cannot find model with the matching brand, model and revision", "", w)
+	// Check that we have a device-serial assertion (the details will have been validated by Decode call)
+	if assertion.Header("type") != "device-serial" {
```
arapulido: We could use the Assertion interface method "Type()" and use the type registry "AssertionType" instead of the header. That way, this won't break if they decide to modify the header.

service/handlers.go
```diff
@@ -107,82 +91,78 @@ func VersionHandler(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusBadRequest)
+		formatSignResponse(false, "error-decode-json", "", err.Error(), nil, w)
 		return
 	}
 
-	// Format the assertions string
-	dataToSign, err := formatAssertion(assertions)
-	if err != nil {
+	// Check that we have a device-serial assertion (the details will have been validated by Decode call)
+	if assertion.Header("type") != "device-serial" {
```
arapulido: We could use the Assertion interface method "Type()" and use the type registry "AssertionType" instead of the header. That way, this won't break if they decide to modify the header.

service/handlers.go
```diff
@@ -107,82 +91,78 @@ func VersionHandler(w http.ResponseWriter, r *http.Request) {
...
 		w.WriteHeader(http.StatusBadRequest)
-		errorMessage := fmt.Sprintf("%v", err)
-		formatSignResponse(false, "error-format-assertions", "", errorMessage, "", w)
+		formatSignResponse(false, "error-decode-assertion", "error-invalid-type", "The assertion type must be 'device-serial'", nil, w)
 		return
 	}
 
-	// Read the private key into a string using the model's signing key
-	privateKey, err := getPrivateKey(model.SigningKey)
+	revision, err := strconv.Atoi(assertion.Header("revision"))
```
arapulido: Checking revision is not needed, it is already done in asserts.Assemble (called at Decode). If Decode has been successful we can assume a valid revision, and it can be retrieved using the Assertion interface function "Revision() int"

arapulido:
Apart from those 2 comments, it looks good. :thumbsup:

service/handlers.go
slimjim777: This check was left as Decode will validate the revision only if it is supplied in the assertion. However, if the revision tag is missing from the assertion, Decode misses that fact and defaults the revision to 0. This will get caught by the FindModel call, which won't find the model.

The TestSignHandlerBadAssertionNoRevision test was there for this scenario, but will need a tweak as my mock FindModel always found the model. I'll change that to do a more realistic check of the supplied parameters.

arapulido: What do you think that misses the point? I guess that if the core library does that, it is because that's how it is supposed to work. Isn't 0 a correct revision? Can you check with Samuele the correct behaviour here? I don't think we should be checking that if they have already done that.

slimjim777: I'm not sure whether 0 is a correct revision or if it's something that's been missed - I'll check with Samuele. I've removed the extra check and altered the mock to do a check of the parameters. The TestSignHandlerBadAssertionNoRevision simulates the case - the outcome is the same (an error is generated), just a different error.

PR #17, opened 2016-05-18
Show models that are linked to inactive keypairs.

slimjim777:


arapulido:
This looks good, merging

PR #19, opened 2016-06-06
Docker compose 1.5.2

swem:
Provide docker-compose files for early evaluation and testing.
Tested on Ubuntu Xenial 16.04. (docker compose 1.5.2, docker 1.10.3)


docker-compose/db/postgres-setup.sql
```diff
@@ -0,0 +1 @@
+CREATE DATABASE identityvault;
```
slimjim777: The settings.yaml file has the default database name as "identity-vault". Let's keep the name consistent.

keystore/private-keys-v0/System/61abf588e52be7a3
```diff
@@ -1,33 +0,0 @@
...
-o2lKqnH3cYOwMypQVzbH8v8nUVyHuBOm/DS8X5TxiSn6f4cdWv49zABDC9SeC5ZjqMkdBOcZBJ4A
-fxIbTGcE+dGNhDQoIi9AVz1evfnFyWqo64/EctDfvCo25aP2mFlzctU11SJr7ifAx90nLrJew0Yn
-p4Ngy8shpllYKC/R53ddTXW7QeDC5ykMmJOWx4rz/KpOX35RG5amNi0aNx8VKWXQFFwYptxZHuvr
-cuGNd75fawlcu6Wklw+dL/z6/PLDYHf1++uHSmibt7ubT9iCtjweGsvEoj05JVq09AXBA4JIr6YX
-N/mv5bawcROq7a6IA5tUYTLkiqLdjAQnnK8H/20w0D9R6J2kcqm9o7RbD+668x2pCj6sy1oHfWUO
-JEpCCleX7u2V8YZDnsV9z0L+7TiTH/TBf+oKiIni7cC66lAQhmDVv0n6t1eFkVyA414ZLnLteSki
-e0OUH7fblJHbbWH50bkW/xTguv9Ri88Tx0F5HiRmF8MPFXBFu69nODQFUL7FshuZfQAJGyGvNvlp
-Lv+9TudssHmrYKV923k/gM/aRKdngzz9YIAUbUqz+mJoOpYSQB61yzGDXuDhPlj8fGHWh+hA9cxz
-x7203qBYiWbfrXtiT5iQzC1UHNqshCgh+qzXLP0ZPUZkqJVYnxTGNEKzgwRZdirRaiSSlfIL4uJ9
-xQ==
```
slimjim777: Removing this test file will generate errors with the tests.

swem: Got it. Add the test file back.

docker-compose/db/postgres-setup.sql
swem: It's seems invalid to have dash "-" on the postgresql database name.
See https://www.postgresql.org/docs/current/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
Only letters, underscores, digits (0-9), or dollar signs ($) is allowed.

swem:
Hi @slimjim777 , I've update pull request according to comments. Could you take a look? Thanks!.

docker-compose/db/postgres-setup.sql
slimjim777: Okay - can you also change the default settings.yaml then. That will keep the names consistent throughout.

swem: Done!.

Issue #43, opened 2017-01-08
Allow configuring settings via system environments


vpetersson:
We're in the early stages of planning to deploy identity-vault. However, one thing that I noticed in terms of a shortcoming is the inability to configure values via environment variables (as far as I can tell).

Our plan is to deploy this in a Kubernetes cluster, and we would like to be able to configure `settings.yaml` as env variables.

I'm happy to contribute the Kubernetes settings/instructions as we go forward.

knitzsche:
Seems like a potentially reasonable feature request. I'll poke someone.

slimjim777:
The application currently sets the config from the command-line parameters and the YAML config file. Adding environment variables to the mix introduces a challenge of which takes precedence. One approach would be to allow the config file command-line parameter to default to blank. So if that is not set, then the application can default to using environment variables. Would that work for Kubernetes?

vpetersson:
@slimjim777 Good point in terms of precedence. From my experience with "Cloud native" applications (i.e. Docker, Kubernetes et al), tend to only have two ways to provide configuration parameters:

 * Config file (which holds some sensible defaults, but no real passwords)
 * Environment variables (which overrides the defaults)

The environment variable always takes precedence over the config file. What we do is normally to have something like this (in Python):

```
[...]
SOME_SETTING = os.getenv('SOME_SETTING', 'some-sensible-default-value')
[...]
```

(I haven't coded anything in Go, but [this StackOverflow post](https://stackoverflow.com/questions/40326540/golang-how-to-assign-default-value-if-env-var-is-empty) seems to implement this.)

By adopting this, we would be able to simplify the Docker example already provided as well by not having to have two different config files.

So I guess the question really is if adding config parameters as command-line parameter is something that makes sense given that there is already a command-line parameter for providing the config file.

It should however be said that it is possible to solve this without it in Kubernetes using [ConfigMaps](https://kubernetes.io/docs/user-guide/configmap/), but that wouldn't be an option for regular Docker deployments.

slimjim777:
The command-line parameters are quite limited (mode and config file path), and they are needed for the snap.

Using the config file and then overriding with environment variable defaults should be fine.

PR #46, opened 2017-01-16
	Optimizes Docker/Docker-compose setup

vpetersson:
* Updates Docker Compose to use v2 format
* Removes unneeded SQL file (automatically done by image)
* Moves configs to environment file (because DRY)
* Generates config file on boot using environment variables

This is a rough workaround, but #43 solves a lot of these workarounds.


vpetersson:
Not sure why Travis is failing here as I only touched the Docker setup. Perhaps it was a temporary error. Could we try to do another re-run of Travis?

slimjim777:
I re-ran the Travis checks and it seemed to work, so I think it was a temporary blip. The issue was related to a file that was not changed. I'm not familiar with Docker, but if these changes are working for you, I'm happy to merge them.

PR #48, opened 2017-02-14
Rand

rmescandon:
Substituted math lib by crypto one when generating random nonces.

### DIFF
--- a/service/utils_test.go
+++ b/service/utils_test.go
@@ -124,3 +124,23 @@ func TestFormatSigningLogResponse(t *testing.T) {
 		t.Errorf("Expected the first fingerprint '%s', got: %s", signingLog[0].Fingerprint, result.SigningLog[0].Fingerprint)
 	}
 }
+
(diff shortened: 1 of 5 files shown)


service/utils_test.go
```diff
@@ -124,3 +124,23 @@ func TestFormatSigningLogResponse(t *testing.T) {
...
+		token, err := GenerateRandomString(1)
+		if err != nil {
+			t.Errorf("Error generating random string: %v", err)
+		}
+		tokens[i] = token
+	}
+
+	// Check that the random strings look unique and valid
+	for i := 1; i < 10; i++ {
+		if tokens[i] == tokens[i-1] {
```
slimjim777: This check looks as though it is only checking adjacent strings for duplication. So if the first and last string were the same, the test would still pass.

It would be better to make sure that none of the strings are duplicated. One approach is to create a map instead of an array:
`var tokens map[string]string`

Then you can check the number of entries instead:
`if len(tokens) < 10:`

PR #54, opened 2017-02-20
Update the installation and deployment instructions

slimjim777:


README.md
```diff
@@ -3,14 +3,50 @@
...
+If you have a Go development environment set up, Go get it:
 
   ```bash
   $ go get github.com/ubuntu-core/identity-vault
   ```
 
-Configure it:
+### Configure it:
 - Install PostgreSQL and create a database.
 - Set up the config file, using ```settings.yaml``` as a guide.
```
rmescandon: Can we be more explicit here for the PostgreSQL complete process?. In my case I had to provide a password for the role and then using it in the connection string. That's not difficult but having the list of commands to execute would be straight. Something like
sudo -u postgres createuser --interactive
sudo -u postgres createdb identityvault
...blah blah

and an example of the settings.yaml connection string to use along with those commands:
datasource: "dbname=identityvault user=theuser password=thepassword"

slimjim777: I've not been explicit here because configuring PostgreSQL is a bit more of a complex area, same with the Apache configuration. There are many ways to set things up, and taking one approach in development may be okay but may not be advisable in production. My concern is that in trying to provide detailed step-by-step instructions, we may lead deployers down the wrong path - especially if they start using dev-oriented instructions in production.

I've placed more of an emphasis on the Juju bundle and the PostgreSQL and Serial Vault charms handle the database creation and configuration themselves.

rmescandon: oki

arapulido:
Looks good now, thanks!

PR #55, opened 2017-02-21
Csrf prevention

rmescandon:
Prevents cross site request forgery issue described in security report. It uses gorilla csrf library [1] which creates random nonces in server used only once from client (react js is the client). Those nonces are put in a frontend template meta header by the server and are sent from client to server using X-CSRF-Token header. Every time the frontend is updated a new nonce is set in meta header. This way an attacker cannot send requests from another client as he doesn't know the generated nonce.
Prevention has been set for POST, PUT and DELETE methods. Idempotent ones are not needed to securize.

[1] https://github.com/gorilla/csrf


rmescandon:
still WIP. Verifying works ok with https

docs/Deployment.md
```diff
@@ -23,7 +23,9 @@ The HA Proxy service provides the load balancing for the additional units.
 
 The Juju bundle does not provide secure system, as all operations are under unencrypted HTTP connections and the Admin service does not providing any authentication.
 The secure connections and authentication need to be handled outside the bundled services. Typically, this will be done by adding front-end web servers that will provide SSL connections
-and authentication for the Admin service. A typical approach would be to use two Apache web servers that provide the SSL connections and authentication for the Admin service.
+and authentication for the Admin service. A typical approach would be to use two Apache web servers that provide the SSL connections and authentication for the Admin service. 
+
+Admin service is protected against cross site request forgery attacks. That means it needs to be accessed through front-end web servers providing the SSL using a valid domain name certificate. It is important accessing the service not using the host IP but the domain name, otherwise requests modifying data won't allowed.
```
slimjim777: *The Admin service

README.md
```diff
@@ -39,6 +40,16 @@ selected service should be accessible on port :8080 or :8081:
  - Signing Service: http://localhost:8080/v1/version
  - Admin Service: http://localhost:8081/
 
+NOTE: Admin service csrf protection enforces accessing the service through HTTPS and using a valid
```
slimjim777: Let's change the wording here:
1. Always capitalise CSRF (it's a title, like HTTPS and SSO).
2. CSRF does not enforce the service through HTTPS, it is just that the cookies are not sent unless HTTPS is used. The workaround for this is setting the environment variable.
3. I don't think you need to have "NOTE:" here, it reads fine without it.

It would be worth reading the docs on the Secure() method as that indicates that it sets the "secure" flag on the cookie, which means that it will only be sent over an HTTPS connection. Disabling this feature with the environment variable ensures that cookies are also sent over HTTP - https://github.com/gorilla/csrf/blob/master/options.go#L45

README.md
```diff
@@ -6,11 +6,12 @@ A Go web service that digitally signs device assertion details.
 The application can be run in two modes: signing service or admin service. Both the web services
 operate under unencrypted HTTP connections, so these should not be exposed to a public network
 as-is. The services should be protected by web server front-end services, such as Apache, that
-provide secure HTTPS connections. Also, the admin service does not include authentication nor
-authorisation, so this service will typically be made available on a restricted network with some
-authentication front-end on the web server e.g. SSO. Typically, the services will only be available
-on a restricted network at a factory, though, with additional security measures, the signing service
-could be made available on a public network.
+provide secure HTTPS connections. Actually, admin service csrf protection forces to be accessed
```
slimjim777: I don't think that adding the CSRF sentence here works as it breaks the flow of the point that is being made about security. I think that the comment that you've added about CSRF later is enough - let's remove that comment from here.

docs/Deployment.md
slimjim777: *will not be allowed.

docs/Deployment.md
```diff
@@ -46,9 +48,11 @@ juju config serial-vault api_keys=Heib2vah2aen3ai
 # Configure the admin service
 #   keystore_secret: part of the key used that is used to encrypt the stored data
 #   api_keys: the key that must be provided in the header of the web service requests
+#	csrf_auth_key: the key to protect server routing from csrf attacks
```
slimjim777: Can you line up "csrf_auth_key" with the other text

docs/Deployment.md
```diff
@@ -46,9 +48,11 @@ juju config serial-vault api_keys=Heib2vah2aen3ai
...
 #   (The keystore_secret and API key must be the same for the two services)
 juju config serial-vault-admin keystore_secret=uXeid2iy1Roo0Io0Beigae3iza5oechu
 juju config serial-vault-admin api_keys=Heib2vah2aen3aid
+juju config serial-vault-admin csrf_auth_key=2E6ZYnVYUfDLRLV/ne8M6v1jyB/376BL9ORnN3Kgb04uSFalr2ygReVsOt0PaGEIRuID10TePBje5xdjIOEjQQ==
```
slimjim777: Not sure about whether we need quotes around the actually csrf_auth_key to make this work (because of '/' characters)

docs/Deployment.md
rmescandon: I've set "32 bytes long key to protect server from cross site request forgery attacks" as text. Is that the one you mean?

Issue #57, opened 2017-03-17
CSRF protection is not secure yet because of cache


rmescandon:
That could happen html page is not refreshed after a non-idempotent operation, so same toked would be used more than once. It is needed to solve this by including token in a previous operation response header before the insert/update/delete operation. That could be an explicit api call to get the token or taking advantage of any existing one

PR #58, opened 2017-03-20
Improve csrf protection

rmescandon:
fixes bug #57 
- Included response header X-CSRF-Token in any backend reply. 
- Updated csrf token in meta header whenever a new respose header is got
- Maintained the old way of setting the meta header by the app handler, as in some cases could happen that the html template is updated after all ajax responses, so that if we remove this, the meta header would be deleted. This way, the last thing happening will update the token.
- Bump version to 1.5


rmescandon:
James proposed a easiest way to perform this

PR #59, opened 2017-03-21
Improve csrf protection take2

rmescandon:
fixes bug #57

 - Updated csrf token before any put, post or delete operation by calling explicit /v1/csrf api resource. Got token from response and injected as new header
 - Bump version to 1.5


service/router.go
```diff
@@ -46,6 +46,9 @@ func AdminRouter(env *Env) *mux.Router {
 	// Start the web service router
 	router := mux.NewRouter()
 
+	// API routes: csrf token
+	router.Handle("/v1/csrf", Middleware(http.HandlerFunc(TokenHandler), env)).Methods("GET")
```
slimjim777: Can we call this /v1/token instead? I'm wondering if it should respond to PUT to make it a little more secure.

webapp/js/models/Ajax.js
```diff
@@ -15,45 +15,59 @@
...
  */
 'use strict'
-var request =require('then-request');
+var request = require('then-request');
 var API_VERSION = '/v1/';
 
 var Ajax = {
+
+	getToken: function() {
+		return this.get('csrf')
```
slimjim777: This endpoint would need to get renamed as well i.e. csrf => token

service/router.go
rmescandon: sounds to me quite generic having "token" in the path.. token of what? a nonce is also a token. makes sense /v1/csrftoken ?

slimjim777: That was the idea - keep it generic, rather than advertising the endpoint as a csrftoken.

PR #60, opened 2017-03-27
System user service

slimjim777:
Create a new system-user service mode to allow the generation of system-user assertions. The service will typically be deployed in a factory rather than the data centre.


webapp-user/src/models/Ajax.js
```diff
@@ -0,0 +1,78 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+var request =require('then-request');
+import axios from 'axios'
```
rmescandon: why using axios? In admin site is not needed

server.go
```diff
@@ -66,6 +66,29 @@ func main() {
...
+		csrfSecureEnv := os.Getenv("CSRF_SECURE")
+		if csrfSecureEnv == "disable" {
+			log.Println("Disable secure flag")
+			csrfSecure = false
+		}
+
+		CSRF := csrf.Protect(
+			[]byte(env.Config.CSRFAuthKey),
+			csrf.Secure(csrfSecure),
+			csrf.CookieName("XSRF-TOKEN"),
```
slimjim777: The default one for Gorilla is starts with and underscore and that seems to cause an issue with retrieving the value. This cookie name is the default for the axios library.

service/handlers_app_user.go
```diff
@@ -0,0 +1,166 @@
...
+type SystemUserResponse struct {
+	Success      bool   `json:"success"`
+	ErrorCode    string `json:"error_code"`
+	ErrorSubcode string `json:"error_subcode"`
+	ErrorMessage string `json:"message"`
+	Assertion    string `json:"assertion"`
+}
+
+// UserIndexHandler is the front page of the web application
+func UserIndexHandler(w http.ResponseWriter, r *http.Request) {
```
slimjim777: The main difference is the name of the template, but that could be refactored to be a parameter.

service/handlers_app_user.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func userRequestToAssertion(user SystemUserRequest, model Model) map[string]interface{} {
+
+	// Create the salt from the keystore secret
+	reg, _ := regexp.Compile("[^A-Za-z0-9]+")
+	baseSalt := reg.ReplaceAllString(Environ.Config.KeyStoreSecret, "")
+
+	// Encrypt the password
+	salt := fmt.Sprintf("$6$%s$", baseSalt)
```
slimjim777: Check the crypt(3) man page: http://man7.org/linux/man-pages/man3/crypt.3.html

snapd only accepts $6 i.e. SHA-512: https://github.com/snapcore/snapd/blob/master/asserts/user.go#L177

service/handlers_app_user.go
```diff
@@ -0,0 +1,166 @@
...
+	password := cryptUser(user.Password, salt)
+
+	// Set the since and end date/times
+	since, err := time.Parse("YYYY-MM-DDThh:mm:ssZ00:00", user.Since)
+	if err != nil {
+		since = time.Now().UTC()
+	}
+	until := since.Add(time.Duration(24*365) * time.Hour)
```
slimjim777: If a model is not specified (it always will be in our case), then the max time is 1 year.
https://github.com/snapcore/snapd/blob/master/asserts/user.go#L250

If used a year as that should be enough for the factory use-case. A const would probably be better.

service/handlers_app_user.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Create the serial assertion header from the serial-request headers
+	headers := map[string]interface{}{
+		"type":              asserts.SystemUserType.Name,
+		"revision":          "1",
```
slimjim777: A const I think

service/handlers_app_user_test.go
```diff
@@ -0,0 +1,150 @@
...
+		{"<invalid\\", 400, false},
+		{generateSystemUserRequestInvalidModel(), 200, false},
+		{generateSystemUserRequestInactiveModel(), 200, false},
+		{generateSystemUserRequestInvalidAssertion(), 200, false},
+		{generateSystemUserRequestInvalidSince(), 200, true},
+	}
+
+	for _, test := range tests {
+		statusCode, result, message := sendSystemUserAssertion(test.data, t)
+		if result != test.expected {
```
slimjim777: Actually, statusCode != text.statusCode

service/middleware.go
```diff
@@ -71,3 +73,16 @@ func Middleware(inner http.Handler, env *Env) http.Handler {
 		inner.ServeHTTP(w, r)
 	})
 }
+
+// CORSMiddleware handles the header options for cross-origin requests (used in development only)
+func CORSMiddleware() func(http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		headers := handlers.AllowedHeaders([]string{"Content-Type", "Content-Length", "Accept-Encoding", "X-CSRF-Token", "X-Requested-With", "Origin"})
+		origins := handlers.AllowedOrigins([]string{"http://localhost:3000"})
+		methods := handlers.AllowedMethods([]string{"GET", "HEAD", "POST", "PUT", "OPTIONS"})
```
slimjim777: With dev mode you can also get OPTIONS - that's generated by a cross-origin request (CORS). That said, I spent a *lot* of time trying to get access from port 3000 working and gave up in the end!

webapp-user/.gitignore
```diff
@@ -0,0 +1,19 @@
+# See https://help.github.com/ignore-files/ for more about ignoring files.
```
slimjim777: There are extra things here that are happening because of the way that create-react-app is being used - particularly related to css files. The css files need to be ignored as they are a part of the build process. So the .gitignore is a part of the create-react-app config.

webapp-user/README.md
```diff
@@ -0,0 +1,56 @@
...
+nvm install lts/*
+
+# Select the version to use
+nvm ls
+nvm use lts/*
+```
+
+- Install the nodejs dependencies
+```bash
+cd webapp-user
```
slimjim777: Not really a typo... just my approach. Often people jump to a particular step and don't read what's above, so this is a "reminder" that they need to be in that directly. Agreed, if the instructions are followed linearly, then that won't be needed.

webapp-user/README.md
```diff
@@ -0,0 +1,56 @@
...
+npm install
+```
+
+## Testing
+To run the tests for an app:
+```bash
+cd webapp-user
+npm test
```
slimjim777: Yes, the test in package.json need to run 'npm run build-css' and then run the tests.

webapp-user/public/index.html
```diff
@@ -0,0 +1,32 @@
...
+      Notice the use of %PUBLIC_URL% in the tag above.
+      It will be replaced with the URL of the `public` folder during the build.
+      Only files inside the `public` folder can be referenced from the HTML.
+
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>Serial User Assertion</title>
+    <meta name="gorilla.csrf.Token" content="not-used-for-dev">
```
slimjim777: Yes, that can be removed

webapp-user/src/App.js
```diff
@@ -0,0 +1,124 @@
...
+import AlertBox from './components/AlertBox'
+import Models from './models/models'
+import Assertion from './models/assertion'
+import {sectionFromHash, handleError} from './models/Utils'
+import createHistory from 'history/createBrowserHistory'
+import './sass/App.css';
+
+const history = createHistory()
+
+class App extends Component {
```
slimjim777: Yes, the "createClass" approach is standard Javascript style (ES5?) and I'm moving towards the ES6 style as that's the future.

This looks like a good intro: https://daveceddia.com/react-es5-createclass-vs-es6-classes/

Here's the react docs on components: https://facebook.github.io/react/docs/components-and-props.html

Just to confuse you even more, they recommend using functions as components wherever possible i.e. whenever they don't need state: https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc

I'm still trying to switch my brain to use functions as components, which is why you aren't seeing too many of them!

webapp-user/src/models/Ajax.js
slimjim777: I started with then-request and switched to axios... I should have remove the then-request dependency.

I was trying to get cookies to work and then-request seems to have a bug related to accessing cookies - since they never appear. I saw in another project that the Design team is using axios, so I thought it may be a better choice.

service/handlers_app_user.go
```diff
@@ -0,0 +1,166 @@
...
+	if err := json.NewEncoder(w).Encode(response); err != nil {
+		logMessage("USER", "signing-assertion", err.Error())
+	}
+}
+
+func userRequestToAssertion(user SystemUserRequest, model Model) map[string]interface{} {
+
+	// Create the salt from the keystore secret
+	reg, _ := regexp.Compile("[^A-Za-z0-9]+")
+	baseSalt := reg.ReplaceAllString(Environ.Config.KeyStoreSecret, "")
```
slimjim777: Do you have any thoughts on using the keystore secret as the baseSalt? Does it compromise the secret? I'm wondering if I should create a hash instead and use that.

slimjim777: @rmescandon that comment was for you!

rmescandon:
LGTM (I assume 32 bits for the salt are enough)

PR #61, opened 2017-04-05
Composite system-user assertion

slimjim777:
Creates a composite system-user assertion through a new web service. The key for signing the system-user assertion is the same as the one used for serial assertions, but it can be different from the one used to sign the model assertion.

This change also includes the start of linking a separate key to a model.


service/handlers_app_user.go
```diff
@@ -111,6 +114,14 @@ func SystemUserAssertionHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Fetch the account assertions from the store
+	accountAssertions, err := fetchAccountAssertions(model)
+	if err != nil {
+		logMessage("USER", "account-assertions", err.Error())
+		formatBooleanResponse(false, "account-assertions", "", "Error retrieving the account assertions from the store", w)
```
slimjim777: Yes, that was my thinking.

rmescandon: In all the other cases where err is being returned have you verified it is not compromising security at all?

slimjim777: I've tried to be consistent with it. The admin UI is probably a bit more lenient as it is aimed at a limited number of sysadmins. The signing service and the system-user service are more "end-user" oriented, so I've tried to limit the raw messages that are sent back.

rmescandon: cool!

PR #62, opened 2017-04-06
Convert the admin webapp to the vanilla-framework

slimjim777:
Convert the admin webapp to the vanilla-framework and fix the react-intl tests.


webapp/js/components/App.js
```diff
@@ -52,23 +55,26 @@ var App = React.createClass({
...
+										<img src="/static/images/logo-ubuntu-white.svg" alt="Ubuntu" />
+										<span>{M({id:"title"})}</span>
+								</div>
 						</div>
-				  </nav>
+
+						<nav role="navigation" className="p-navigation__nav">
+							<span className="u-off-screen"><a href="#navigation">Jump to site nav</a></span>
+							<Navigation />
+								{/*<form id="language-form" className="header-search">*/}
```
rmescandon: Why not rendering langs change option?

slimjim777: We only have an English translation at the moment, so I've been wanting to remove the option for a while. If/when we get the Chinese translation done, we can re-enable this - all the other internationalization still works.

rmescandon:
approved!

PR #65, opened 2017-04-12
Use the system-user key for account key.

slimjim777:
Change labels for the keys.
Show the version in the footer.


rmescandon:
All perfect. LGTM

PR #66, opened 2017-04-18
Cache account keys

slimjim777:
Account keys are retrieved in a separate account-assertion command from the store and cached in the database. The system-user assertion uses the cached account keys rather than trying to get them from the store.


service/keypairmanager.go
```diff
@@ -63,6 +69,16 @@ func (db *DB) CreateKeypairTable() error {
...
 }
 
+// AlterKeypairTable adds extra fields to an existing keypair database table
+func (db *DB) AlterKeypairTable() error {
+	_, err := db.Exec(alterKeypairAddAssertion)
+	if err != nil {
+		// Ignore error as the field may already be added
+		return nil
+	}
+	return err
```
rmescandon: ??? weird method. One improvement could be warning about the error. But in any case, you are always returning nil, no matter whether there is an error or not. Why not:
func (db *DB) AlterKeypairTable() {
   db.Exec(alterKeypairAddAssertion)
}

slimjim777: The "Alter" methods are the ones that most likely will get additions in the future, so I used that structure. Probably unnecessary, as you say.

PR #67, opened 2017-04-21
Display the account and account key assertions in the admin interface

slimjim777:
Display the status of the assertions for keypairs, depending on whether they are used for signing system-users or not.

(This version does not handle manual upload of keys)


webapp/js/components/AccountList.js
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+
+    renderAccountKeys() {
+        if (this.state.keypairs.length > 0) {
+            return (
+                <table>
+                <thead>
+                    <tr>
+                    <th>{T('key-id')}</th><th>{T('status')}</th>
```
rmescandon: Maybe having "assertion" as label instead of status would be better. In case all is ok, the assertion is shown as value for the status column, what looks a bit weird

webapp/.babelrc
```diff
@@ -1 +1,4 @@
-{ "presets": ["es2015", "react", "stage-0"] }
+{ 
+    "presets": ["es2015", "react", "stage-0", "airbnb"],
```
slimjim777: You get an AirBnB voucher with each new deployment ;)
The tests were failing for new ES6 code and I wanted to bring this inline with the testing approach in create-react-app. That uses enzyme for testing, which is an airbnb  project.

Babel is a pre-processor that converts ES6 style code into native ES5 Javascript. This is so that projects can use ES6 now without waiting years for browsers to handle ES6 natively. There are processors for react (converting JSX) and airbnb is doing some magic... it fixed the issues that I had and allowed enzyme tests to work.

webapp/js/app.js
```diff
@@ -27,6 +27,7 @@ var ModelList = require('./components/ModelList');
 var ModelEdit = require('./components/ModelEdit');
 var KeypairAdd = require('./components/KeypairAdd');
 var SigningLog = require('./components/SigningLog');
+import AccountList from './components/AccountList';
```
slimjim777: Yes, the new-style components need an import. We can switch things over to import completely. It will take a bit more work to bring everything to ES6. By then, we'll probably find that ES7 has changed all the rules again!

webapp/js/components/AccountList.js
slimjim777: Agreed

rmescandon:
LGTM

PR #68, opened 2017-04-26
Account key upload

slimjim777:


webapp/js/components/Utils.js
```diff
@@ -23,3 +23,35 @@ export function T(message) {
     return msg
 }
 
+export function isHTTPError(response) {
+    // Check for an HTTP error
+    if ((response.statusCode >= 300) && (response.statusCode != 400)) {
```
rmescandon: Not sure that 3xx codes can be considered errors, but 400 is indeed an error. What's the purpose of this method? Maybe the name or the conditions to return true/false should be changed?

rmescandon:
Btw, aspect of web is correct. The ugly view was happening only in my laptop

webapp/js/components/Utils.js
slimjim777: The API returns an error 400 for "expected errors" i.e. ones that have been caught and returned as a JSON response. So, the normal responses will always be 200 or 400. Anything else is something unexpected happening at the server and it is likely that the response won't be JSON. So this is to handle the unexpected errors gracefully.

rmescandon: so isUnexpectedHTTPError() isUnmanagedHTTPError()... names for the function?

slimjim777: I've found that the function is not used. It's imported in AccountForm.js, but has been replaced with parseError. I'll merge for now and remove it in the next PR, which is in progress.

PR #70, opened 2017-05-02
New snap to install all the Serial Vault Services.

slimjim777:
New script to handle the database installation, initialization and configuration for the services.


serial-vault-services-snap/install.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
+
```
rmescandon: It would be nice having some check to see if current user is root and exit if not. 

if [ "$EUID" -ne 0 ]
  then echo "Please run as root"
  exit
fi

or 

if [ $(id -u) -ne 0 ]
  then echo "Please run as root"
  exit
fi

serial-vault-services-snap/install.sh
```diff
@@ -0,0 +1,62 @@
...
+adduser --disabled-login --disabled-password --gecos "" serialvault
+
+# Install postgresql
```
rmescandon: sudo ./install.sh reports the error

pg_ctl: directory "/home/serialvault/snap/postgresql96/common/data" is not a database cluster directory

and after that the services are not started and installation is not ok at all.
I've been trying to execute it with another user (postgres) with same result, even trying to doing it by hand.
In my environment there is no previous postgresql installation.

rmescandon:
Now I get https://pastebin.canonical.com/187512/

serial-vault-services-snap/bin/serviceinit.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/sh
+
+# Initialize and start the database service
+echo "Initialize the database"
+$SNAP/usr/bin/wrapper-initialize initdb
+$SNAP/usr/bin/wrapper-pg_ctl -D $SNAP_USER_COMMON/data -l $SNAP_USER_COMMON/logs/logfile start
```
rmescandon: maybe it would be good to verify if this operation results in an error due to locale or even set directly LC_ALL=en_US.UTF-8

serial-vault-services-snap/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+## Install the Serial Vault Services
+The Serial Vault Services need to have a PostgreSQL service installed and configured. The 
+installation script simplifies this process by handling the full installation process.
+
+```bash
+cd serial-vault-services-snap  # Make sure you are in the correct directory
+sudo snap install --dangerous serial-vault-services_1.5_amd64.snap
+```
```
slimjim777: I've added a comment about 'sudo' and used the wildcard.

serial-vault-services-snap/bin/serviceinit.sh
slimjim777: I've added this to the serviceinit script. Ideally I'd like the serviceinit to run automatically (as a hook?), but I'm not sure if that will be possible.

rmescandon: sorry, I forgot to reply to this. There is indeed a hook to be executed when snap configuration changes. That includes when first install or in any upgrade. You could use it to execute serviceinit. See https://github.com/snapcore/snapd/wiki/hooks
Maybe implement this in another PR, better

rmescandon:
In root folder README.md there is a section "install using the snap package" that doesn't fit with the one in the subfolder serial-vault-services-snap. Can you please, redirect from there to this last in that README?

rmescandon:
LGTM
I wonder how can deployer get the info of keystore secret or api key if using a hook to start the services. Maybe "config" operation could provide that info if not got any param?

slimjim777:
There are two snaps in this project:
* root folder: just installs one service
* serial-vault-services-snap: installs all services including Postgres and Apache

The first snap is the one that would be normally used for deployment - that snap is published in the store and there is a juju charm and bundle that uses it. The root README instructions are focused on the typical deployment.

This second snap "bundle" is an all-in-one serial vault that can be used to simplify deployment on a restricted factory network. This snap is not published in the store and would be built and installed by the company that needs it. This is really just a convenience solution for a factory network, not the recommended solution for most customers.

slimjim777:
I'll add a bug about getting the config.

rmescandon:
Thanks for the explanation. I was not sure about the difference and If both were needed.

PR #72, opened 2017-05-12
Default to HTTP and allow SSL via the 'enable-https' command

slimjim777:


serial-vault-services-snap/README.md
```diff
@@ -51,12 +51,12 @@ Signing Service : http://localhost/signing/v1/version
 
 
 ## Set-up Apache and SSL
-Apache is configured to use a self-signed certificate, which will cause a browser warning
-for the certificate. It is possible to supply your own certificate using the ```enable-https``` command.
-
+Apache is configured to use HTTP by default. It is possible to use HTTPS by generating a self-signed certificate or
+supplying to supply your own certificate using the ```enable-https``` command.
```
rmescandon: 'supplying to supply' sounds like one or the other ;)

PR #73, opened 2017-05-15
Update the basic snap to add the new system-user mode

slimjim777:


rmescandon:
As discussed on irc, this branch can be landed as is. No version bump is needed and other changes can be performed in other PRs
LGTM

PR #76, opened 2017-06-27
Add Ubuntu SSO authentication with OpenID

slimjim777:


usso/jwt.go
```diff
@@ -0,0 +1,99 @@
...
+
+// NewJWTToken creates a new JWT from the verified OpenID response
+func NewJWTToken(resp *openid.Response) (string, error) {
+	token := jwt.New(jwt.SigningMethodHS256)
+
+	token.Claims[ClaimsUsername] = resp.SReg["nickname"]
+	token.Claims[ClaimsName] = resp.SReg["fullname"]
+	token.Claims[ClaimsEmail] = resp.SReg["email"]
+	token.Claims[ClaimsIdentity] = resp.ID
+	token.Claims["exp"] = time.Now().Add(time.Hour * 24).Unix()
```
rmescandon: why exp is not using a constant like the others?

usso/jwt_test.go
```diff
@@ -0,0 +1,100 @@
...
+	test1 := testJWT{
+		resp:     openid.Response{ID: "id", Teams: []string{"teamone", "team2"}},
+		expected: []string{"", "", ""},
+	}
+	test2 := testJWT{
+		resp:     openid.Response{ID: "id", Teams: []string{"teamone", "team2"}, SReg: map[string]string{"nickname": "jwt"}},
+		expected: []string{"jwt", "", ""},
+	}
+	test3 := testJWT{
+		resp:     openid.Response{ID: "id", Teams: []string{"teamone", "team2"}, SReg: map[string]string{"nickname": "jwt", "email": "jwt@example.com", "fullname": "John W Thompson"}},
```
rmescandon: just a matter of format: these lines are too long. Can you break the line like resp:     
                openid.Response{
			ID: "id", 
			Teams: []string{"teamone", "team2"}, 
			SReg: map[string]string{"nickname": "jwt", "email": "jwt@example.com", "fullname": "John W Thompson"}},
                }

usso/jwt_test.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func expectedToken(t *testing.T, jwtToken string, resp *openid.Response, username, email, name string) {
+	token, err := VerifyJWT(jwtToken)
+	if err != nil {
+		t.Errorf("Error validating JWT: %v", err)
+	}
+
+	if token.Claims[ClaimsIdentity] != resp.ID {
+		t.Errorf("JWT ID does not match: %v", token.Claims[ClaimsIdentity])
```
rmescandon: btw. In other projects (snapd, wifi-ap, ...) are using https://labix.org/gocheck package for testing. That makes code smaller and assertions in only one line. Things like this are resumed in something like:
c.Assert(token.Claims[ClaimsIdentity], check.Equals, resp.ID)

a random example line:
https://github.com/snapcore/snapd/blob/master/cmd/snap/cmd_auto_import_test.go#L56

In a future refactor PR we can migrate to this. In wifi-connect i tried it and works fine (and skips writing the specific t.Errorf())

usso/openid.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/juju/usso/openid"
+)
+
+var (
+	teams    = "ce-web-logs,canonical"
+	required = "email,fullname,nickname"
+	optional = ""
+)
+
+var client = openid.NewClient(usso.ProductionUbuntuSSOServer, &datastore.OpenidNonceStore, nil)
```
rmescandon: I guess that client already knows the endpoint to be redirected (Location header) for open id authentication.
In that case, what happens if that url changes? (a preproduction openid environment, etc..). Or is that configurable somewhere that I've missed?
(btw, in Multitenancy doc maybe it would be nice having a schema of data exchange in openid login phase)

usso/openid.go
```diff
@@ -0,0 +1,103 @@
...
+
+// verify is used to perform the OpenID verification of the login
+// response. This is declared as a variable so it can be overridden for
+// testing.
+var verify = client.Verify
```
rmescandon: This is valid, yes, but in cases like this where there is already a struct whose methods you are mocking, maybe it is interesting to have an interface + mocked struct. 
What i made in wifi-connect is having 
var client <theInterface> = openid.NewClient(...
and assign another value to client when testing.

In any case, as here you are still only using one method, leave as is

usso/jwt_test.go
```diff
@@ -0,0 +1,100 @@
...
+		if err != nil {
+			t.Errorf("Error creating JWT: %v", err)
+		}
+
+		expectedToken(t, jwtToken, &r.resp, r.expected[0], r.expected[1], r.expected[2])
+
+	}
+}
+
+func expectedToken(t *testing.T, jwtToken string, resp *openid.Response, username, email, name string) {
```
rmescandon: not sure if in this method or outside, but I miss some test to verify that the token is not expired. Also that the expiration time is more than and less than ... whatever.
(maybe VerifyJWT() is doing this, not sure)

account/account_test.go
```diff
@@ -23,34 +23,35 @@ import (
...
-	"github.com/CanonicalLtd/serial-vault/service"
+	"github.com/CanonicalLtd/serial-vault/config"
+	"github.com/CanonicalLtd/serial-vault/datastore"
 	"github.com/snapcore/snapd/asserts"
 )
 
 func TestCacheAccountAssertions(t *testing.T) {
 	// Mock the database
-	mockDB := service.MockDB{}
-	config := service.ConfigSettings{KeyStoreType: "filesystem", KeyStorePath: "../keystore", KeyStoreSecret: "secret code to encrypt the auth-key hash"}
```
slimjim777: The tests are still in service.utils - I'll move them

cmd/account-assertions/main.go
```diff
@@ -23,22 +23,23 @@ import (
 	"log"
 
 	"github.com/CanonicalLtd/serial-vault/account"
-	"github.com/CanonicalLtd/serial-vault/service"
+	"github.com/CanonicalLtd/serial-vault/config"
+	"github.com/CanonicalLtd/serial-vault/datastore"
 )
 
 func main() {
```
slimjim777: It just started there and has never been moved. I've been wary of messing up the deployed snap and the docker script has the location coded there. This may be a good time to move it, but it will result in more changes i.e. to the snap files... brace yourself!

datastore/database.go
```diff
@@ -67,15 +67,31 @@ type Datastore interface {
...
+	Config    config.Settings
+	DB        Datastore
+	KeypairDB *KeypairDatabase
+}
+
+// Environ contains the parsed config file settings.
+var Environ *Env
+
+// OpenidNonceStore contains the database nonce store for Openid
+var OpenidNonceStore PgNonceStore
```
slimjim777: The way the var is used causes a nil pointer exception at start-up if it is part of Environ.

datastore/dbkeypairoperator.go
```diff
@@ -17,13 +17,15 @@
...
-package service
+package datastore
 
 import (
 	"crypto/hmac"
 	"crypto/sha256"
 	"encoding/base64"
 	"log"
+
+	"github.com/CanonicalLtd/serial-vault/utils"
```
slimjim777: Okay - changed to crypt and random (rand clashes with the standard library)

datastore/mock_database.go
```diff
@@ -484,3 +494,13 @@ func (mdb *ErrorMockDB) CreateDeviceNonce() (DeviceNonce, error) {
 func (mdb *ErrorMockDB) ValidateDeviceNonce(nonce string) error {
 	return errors.New("MOCK error validating a nonce")
 }
+
+// CreateOpenidNonceTable database mock
+func (mdb *ErrorMockDB) CreateOpenidNonceTable() error {
+	return nil
```
slimjim777: If a CreateTable fails then there is an issue with the database and the whole service will fail (as it gets run on service restart). Not sure if it's really worth the extra effort.

That makes sense, but it means changes to the snaps as well :)

datastore/openidmanager.go
```diff
@@ -0,0 +1,102 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2017-2018 Canonical Ltd
```
slimjim777: I've never understood what the year range means in the copyright files!

datastore/openidmanager.go
```diff
@@ -0,0 +1,102 @@
...
+`
+
+// Indexes
+const createOpenidNonceIndexSQL = "CREATE UNIQUE INDEX IF NOT EXISTS nonce_endpoint_idx ON openidnonce (nonce, endpoint)"
+
+// Queries
+const createOpenidNonceSQL = "INSERT INTO openidnonce (nonce, endpoint, timestamp) VALUES ($1, $2, $3)"
+const deleteExpiredOpenidNonceSQL = "DELETE FROM openidnonce where timestamp<$1"
+
+// OpenidNonce holds the details of the nonce, combining a timestamp and random text
```
slimjim777: There are checks done to stop replay attacks and that combines the nonce string and the endpoint. I'll add a link to the docs... not that they are very clear!

datastore/openidmanager.go
```diff
@@ -0,0 +1,102 @@
...
+	// Create the index
+	_, err = db.Exec(createOpenidNonceIndexSQL)
+	return err
+}
+
+// CreateOpenidNonce stores a new nonce entry
+func (db *DB) CreateOpenidNonce(nonce OpenidNonce) error {
+
+	// Delete the expired nonces
+	err := db.deleteExpiredOpenidNonces()
```
slimjim777: Agreed

datastore/openidnoncestore.go
```diff
@@ -0,0 +1,77 @@
...
+	// following restrictions:
+	//
+	// + All times must be in the UTC timezone, indicated with a "Z".
+	//
+	// + No fractional seconds are allowed
+	//
+	// For example: 2005-05-15T17:11:51ZUNIQUE
+
+	if len(nonce) < 20 {
+		return errgo.Newf("%q does not contain a valid timestamp", nonce)
```
slimjim777: I've based this method on other implementations and all the messages are vague e.g. "invalid nonce"! This should come directly from Ubuntu SSO, so anything unexpected is highly suspicious, so being vague is fine.

get-deps.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -eu
```
slimjim777: I've removed it - it was a leftover from trying to fix Travis

service/handlers_test.go
```diff
@@ -382,44 +385,21 @@ func TestSignHandlerErrorKeyStore(t *testing.T) {
...
 
 	result, _ := sendRequestVersion(t, "GET", "/v1/version", nil)
 
-	if result.Version != Environ.Config.Version {
-		t.Errorf("Incorrect version returned. Expected '%s' got: %v", Environ.Config.Version, result.Version)
-	}
-
-}
-
-func TestVersionHandlerNilEnviron(t *testing.T) {
```
slimjim777: This was testing a local env variable and the package Environ variable being set when it was nil. The local env variable has been removed and Environ is always initialized, so it doesn't make sense any more.

service/middleware.go
```diff
@@ -86,3 +81,45 @@ func CORSMiddleware() func(http.Handler) http.Handler {
...
+		if !datastore.Environ.Config.EnableUserAuth {
+			protected.ServeHTTP(w, r)
+			return
+		}
+
+		// Get the JWT from the header or cookie
+		jwtToken, err := usso.JWTExtractor(r)
+		if err != nil {
+			log.Println(err.Error())
+			// TODO: don't leave pages unprotected when there is no token
```
slimjim777: Yes - I imagine that it would redirect to / with an error message, but I'm open to suggestions... when we come to that.

usso/jwt.go
slimjim777: exp is one of the standard values of a JWT, whereas the others are custom. The jwt library used to have a constant but that has been removed and there are no examples that match the current code: https://godoc.org/github.com/dgrijalva/jwt-go#example-New--Hmac

I think I'll add our own constant for now, but hopefully the jwt library will improve their examples.

usso/jwt_test.go
slimjim777: Done

usso/jwt_test.go
slimjim777: The token expiry and validity (if it uses a future time) is handled by the parser library in VerifyJWT. So we just need to check the custom data that is expect.

usso/jwt_test.go
slimjim777: Yes - worth filing a card for that

usso/openid.go
slimjim777: The library we are using is specifically for USSO and it handles the staging USSO site and production. However, I can't see us ever using the staging site - on our staging deployment we'd still use the production USSO site. The library will be kept up to date with USSO, so all we'd need to do is use the latest library and recompile.

I'll document the claims in the JWT. The USSO library abstracts away the raw responses from openid so I don't think it's worth repeating those.

usso/openid.go
slimjim777: I'm not really sure that I understand this in this particular case. If you look at the implementation of the openid library, they are using the same approach: https://github.com/juju/usso/blob/master/openid/openid.go#L182

The reason to only mock Verify is that it does a callback to the USSO service to check that the details are from the real server. So this is just mocking that callback.

datastore/openidmanager.go
rmescandon: iirc it is the range of years the development of that file took

usso/jwt_test.go
rmescandon: I've added a card for this https://trello.com/c/hg4TXDAh/367-serial-vault-migrate-tests-to-gocheck

PR #77, opened 2017-07-03
User tables

rmescandon:
Created tables for storing users information and its relationship with accounts (n-m)
Updated Database interface to include User operations
Updated database mocks to implement new Database methods
Updated createdb command to include new tables creation.

This PR does not include any use of data in those new tables, only infrastructure management


datastore/usermanager.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+const createUserTableSQL = `
+	CREATE TABLE IF NOT EXISTS userinfo (
+		id               serial primary key not null,
+		username         varchar(200) not null unique,
+		openid_identity  text not null,
+		name             varchar(200),
+		email            varchar(255) not null,
+		role             int not null
```
slimjim777: I think that 'role' is a reserved word - Postgres has stolen all the good names!

rmescandon: couriously, there are no errors in sql operations for that table.... but you're right, we better use another name. Suggestions?...
Btw, I finally used 'userinfo' for the table, as you can see. Sounded weird to me having table named as 'profile' and then the related struct as User

datastore/mock_database.go
```diff
@@ -315,6 +316,89 @@ func (mdb *MockDB) CreateOpenidNonce(nonce OpenidNonce) error {
...
+		OpenIDIdentity: "https://login.ubuntu.com/+id/Abcyssfmr",
+		Name:           "Rigoberto Picaporte",
+		Email:          "rigoberto.picaporte@ubuntu.com",
+		Role:           Standard})
+	users = append(users, User{
+		ID:             2,
+		Username:       "user2",
+		OpenIDIdentity: "https://login.ubuntu.com/+id/Abcyssfmr",
+		Name:           "Nancy Reagan",
+		Email:          "nancy.reagan@usa.gov",
```
slimjim777: It's an honour to have Nancy assist with testing the serial-vault. Does she need a unique OpenID identity?

datastore/usermanager.go
slimjim777: I found the same with table names - they should be only alphanumeric but it accepts 'serial-vault' when it shouldn't.

rmescandon: renamed to 'userrole'. Let me know if you prefer another name

datastore/mock_database.go
rmescandon: Of course. Such a VIP needs a different OpenIDIdentity than this comic guy http://cloud10.todocoleccion.online/tebeos-bruguera/tc/2015/06/02/11/49631384.jpg

PR #79, opened 2017-07-04
Display the user interface based on the user's permissions

slimjim777:


webapp/js/components/Utils.js
```diff
@@ -54,3 +72,82 @@ var getQueryString = function ( field, url ) {
...
+    if (!token.role) return false
+
+    if (token.role >= Role.Admin) {
+        return true
+    } else {
+        return false
+    }
+}
+
+export function isUserSuperuser(token) {
```
rmescandon: why 3 methods for the same thing? If that's just for clarity, why not better:
export function isUserAdmin(token) {
   return isUser(admin, token)
}
export function isUserStandard(token) {
   return isUser(stadandard, token)
}
export function isUserSuperuser(token) {
   return isUser(superuser, token)
}
export function isUser(role, token) {
    if (!token) return false
    if (!token.role) return false

    if (token.role >= role) {
        return true
    } else {
        return false
    }
}

PR #80, opened 2017-07-05
Allow access to the admin interface if user authentication is disabled

slimjim777:


slimjim777:
At the moment there are no superuser pages, but the superuser will be able to create new accounts and users. So all the user management will need to be added to the UI, but won't be displayed when auth is disabled.

PR #82, opened 2017-07-07
Filter the list models API based on the user

slimjim777:
Refactor the JWT validation so that the token can be used by protected handlers.

Handle filtering of the list models API based on the user.

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -90,6 +90,9 @@ func TokenHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
 	w.Header().Set("X-CSRF-Token", csrf.Token(r))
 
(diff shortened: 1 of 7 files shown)


service/handlers.go
```diff
@@ -90,6 +90,9 @@ func TokenHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
 	w.Header().Set("X-CSRF-Token", csrf.Token(r))
 
+	// Check the JWT and return it in the authorization header, if valid
+	JWTCheck(w, r)
```
rmescandon: What happens if this returns an error? Shouldn't this return some kind of http error in that case?

slimjim777: No - this is just used by the UI. So the JWT will be set correctly or not. If not, the UI will set up nil permissions.

PR #83, opened 2017-07-07
Handle when the user unticks mandatory fields in the OpenID login

slimjim777:


usso/openid.go
```diff
@@ -78,15 +77,18 @@ func LoginHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	// get form username and get from datastore the User
+	// Check we have the mandatory fields in the OpenID response
 	username := r.Form.Get("openid.sreg.nickname")
-	if len(username) == 0 {
-		log.Println("Got no 'openid.sreg.nickname' in response params")
-		replyHTTPError(w, http.StatusBadRequest, errgo.New("OpenID response has not valid format"))
+	fullname := r.Form.Get("openid.sreg.fullname")
+	email := r.Form.Get("openid.sreg.email")
+	if len(username) == 0 || len(fullname) == 0 || len(email) == 0 {
```
rmescandon: I think (correct me if I'm wrong) that strictly needed can only be the nickname. And in case fullname is empty, we could fill User.Name with nickname content. Email i guess that could be empty if not received

slimjim777: The nickname is used in the backend and the fullname is used in the front-end, so both are needed. Strictly, the email isn't needed but we may need it in the future if we want to notify the user of something. Maybe that can be removed for now.

rmescandon: But the fullname in the frontend is used to identify if there is an authenticated user (evaluating if not null) and to show it in the navigation bar button, right?. In those both cases it doesn't matter if its value is the nickname or the full name; the important is having any value (is there any other use that needs the real full name and the nickname replacement is not valid?). The idea behind this is letting the user being authenticated even if he doesn't want to share more than the nickname

PR #84, opened 2017-07-11
Add API permissions to the model endpoints

slimjim777:


service/utils.go
```diff
@@ -198,3 +199,23 @@ func validateModelName(name string) error {
 
 	return nil
 }
+
+// checkUserPermissions retrieves the user from the JWT.
+// The user will be restricted by the accounts the username can access and their role.
+// If user authentication is turned off, the JWT will irrelevant.
+func checkUserPermissions(w http.ResponseWriter, r *http.Request) (string, error) {
```
rmescandon: still thinking for a method name...
getAllowedUser(), getAuthUser()...
checkUserPermissions() sounds to me that you provide the permissions that are checked against something. Here is the token what is checked and from there it is extracted the allowed user (or no user if * is allowed)

I could accept checkUser() even as valid

slimjim777: I think I've added a check that we have an admin user in the next PR :)

datastore/usermanager.go
```diff
@@ -209,3 +217,32 @@ func rowsToUsers(rows *sql.Rows) ([]User, error) {
 
 	return users, nil
 }
+
+func (db *DB) checkUserPermissions(username string) int {
```
slimjim777: This gets changed in the next PR! Can we let this through here and I'll take the naming into account there.

rmescandon: sure.

PR #85, opened 2017-07-13
Add permissions and filtering to API endpoints

slimjim777:
* Keypairs
* Signing Log
* Accounts


service/utils.go
```diff
@@ -217,5 +217,13 @@ func checkUserPermissions(w http.ResponseWriter, r *http.Request) (string, error
 	}
 
 	// Get the user from the token
-	return token.Claims[usso.ClaimsUsername].(string), nil
+	username := token.Claims[usso.ClaimsUsername].(string)
+
+	// Check that the role is at least an Admin
+	role := datastore.Environ.DB.RoleForUser(username)
+	if role < datastore.Admin {
```
rmescandon: retricting to Admin or more is another reason to modify method name. Though now I'm not sure what name is suitable, because this does some different things...
YoGimmeDaUserCrapBro()
AllowUser() 
checkUserPermissions()

ok, accepted checkUserPermissions, BUT...
modify comments of the method, explaining better than the method validates Admin and Superusers, and explaining the return value and what implies returning an empty username

datastore/database.go
```diff
@@ -53,17 +53,16 @@ type Datastore interface {
 	CreateSigningLogTable() error
 	CheckForDuplicate(signLog *SigningLog) (bool, int, error)
 	CreateSigningLog(signLog SigningLog) error
-	ListSigningLog() ([]SigningLog, error)
-	DeleteSigningLog(signingLog SigningLog) (string, error)
```
slimjim777: This was initially included as it was not allowed to get a serial assertion for the same device more than once. This allowed developers to delete the log entry and reflash the device. Now you can actually get a serial assertion for the same device multiple times and you get a bumped revision number. So there is no longer a need to delete a signing log entry.

The feature was removed from the UI a long time ago... this is cleaning up the backend.

service/handler_accounts.go
```diff
@@ -48,7 +48,14 @@ type AssertionRequest struct {
 // AccountsHandler is the API method to list the account assertions
 func AccountsHandler(w http.ResponseWriter, r *http.Request) {
 
-	accounts, err := datastore.Environ.DB.ListAccounts()
+	// Get the user from the JWT
+	username, err := checkUserPermissions(w, r)
```
slimjim777: No - as I mentioned in the previous PR, I've now added a permissions check into that method. So the name now  makes sense.

PR #90, opened 2017-07-18
User management backend

rmescandon:
Db, handling and routing implementation to expose users management backend.
All methods are only allowed to Superusers
API unit tests


service/utils.go
```diff
@@ -178,35 +200,58 @@ func logMessage(method, code, reason string) {
...
+
+func validateUsername(username string) error {
+	return validateSyntax("Username", username, validUsernamePattern)
+}
+
+func validateUserRole(role int) error {
+	return role == datastore.Standard || role == datastore.Admin || role == datastore.Superuser
+}
+
+func validateSyntax(fieldName, fieldValue, pattern string) error {
```
slimjim777: Nice!

PR #95, opened 2017-07-27
Fix JWT claim properties 

pvanloo:
The jwt-go package has upgraded to version 3 and has changed the way claim properties are accessed from tokens

### DIFF
--- a/dependencies.tsv
+++ b/dependencies.tsv
@@ -1,5 +1,5 @@
 github.com/cheggaaa/pb	git	0817e3a1f8de9e3c78b159699b3c07d53e24a963	2017-03-23T12:38:18Z
-github.com/dgrijalva/jwt-go	git	afef698c326bfd906b11659432544e5aae441d44	2015-12-30T18:00:00Z
+github.com/dgrijalva/jwt-go	git	a539ee1a749a2b895533f979515ac7e6e0f5b650	2017-06-08T02:51:49Z
 github.com/gorilla/context	git	08b5f424b9271eedf6f9f0ce86cb9396ed337a42	2016-08-17T18:46:32Z
 github.com/gorilla/csrf	git	69581736821c33d85bbf378f42f6ad864dbd85de	2016-11-22T16:45:00Z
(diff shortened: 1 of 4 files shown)


dependencies.tsv
```diff
@@ -1,5 +1,5 @@
 github.com/cheggaaa/pb	git	0817e3a1f8de9e3c78b159699b3c07d53e24a963	2017-03-23T12:38:18Z
-github.com/dgrijalva/jwt-go	git	afef698c326bfd906b11659432544e5aae441d44	2015-12-30T18:00:00Z
+github.com/dgrijalva/jwt-go	git	a539ee1a749a2b895533f979515ac7e6e0f5b650	2017-06-08T02:51:49Z
```
slimjim777: You appear to be using the head version of the repo rather than the released version. I'd prefer to stick to the tagged release for version 3.0.0:
`github.com/dgrijalva/jwt-go     git     d2709f9f1f31ebcda9651b03077758c1f3a0018c        2016-06-16T19:15:56Z`

service/utils.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/CanonicalLtd/serial-vault/datastore"
 	"github.com/CanonicalLtd/serial-vault/usso"
 	"github.com/snapcore/snapd/asserts"
+	"github.com/dgrijalva/jwt-go"
```
slimjim777: The build process is giving a gofmt error. It seems to need the imports sorted alphabetically i.e. move the jwt-go import up one line.

usso/jwt_test.go
```diff
@@ -25,6 +25,7 @@ import (
 	"testing"
 
 	"github.com/juju/usso/openid"
+	"github.com/dgrijalva/jwt-go"
```
slimjim777: The build process is giving a gofmt error. It seems to need the imports sorted alphabetically i.e. move the jwt-go import up one line.

pvanloo:
Thanks for reviewing, I've made the requested changes.

PR #105, opened 2017-08-04
Add api-key to the UI for the model and validate and default the field

### LINKED ISSUE #102: Add API key to the model UI
The API key has been moved from the config file to a model. The UI needs to be changed so that it displays and allows the API key to be changed.

slimjim777:
Closes #102


datastore/modelmanager.go
```diff
@@ -462,3 +474,21 @@ func (db *DB) checkBoolQuery(row *sql.Row) bool {
...
+func (db *DB) defaultAPIKey(model *Model) error {
+	// Remove all whitespace from the API key
+	model.APIKey = strings.Replace(model.APIKey, " ", "", -1)
+
+	// Check we have a minimum API key size
+	if len(model.APIKey) > minAPIKeyLength {
+		return nil
+	}
+
+	apiKey, err := db.generateAPIKey()
```
rmescandon: why generateAPIKey() is a db method and not a standalone one?
If it can be a standalone one, also defaultAPIKey() could be

datastore/modelmanager.go
```diff
@@ -462,3 +474,21 @@ func (db *DB) checkBoolQuery(row *sql.Row) bool {
 
 	return found
 }
+
+// defaultAPIKey creates a default API key if the field is empty
+func (db *DB) defaultAPIKey(model *Model) error {
```
rmescandon: Not sure if I like the name and signature of the method.
I'd prefer this style:
func (db *DB) buildValidOrDefaultAPIKey(originalKey string) (string, err) {..}

and use it like:
model.APIKey, err = buildValidOrDefaultAPIKey(model.APIKey)

In any case, if you want to directly modify model (not recommended style in certain book I read recently and that can become your nightmare), at least modify the method name to indicate that it replaces whitespaces, checks minimum size and generate a new api key (difficult name, but no matter if it is really long)

PR #112, opened 2017-08-08
Superuser accounts permissions

### LINKED ISSUE #101: Superuser cannot see all accounts
The superuser should be able to see all accounts, but currently can only see the ones assigned to the current user. This also happens with models: current user cannot see models whose brand is not an account related to him

rmescandon:
* Moved auth logic outside database layer
* Removed any condition involving a role in database sql statements
* Created a datastore.User object with request auth data and evaluate permissions in every handler operation using that object
* Updated existing tests to use this auth
* Added additional tests to handlers' in order to test login users with less permissions than required

Fixes #101


service/handlers_keypairs.go
```diff
@@ -64,15 +63,34 @@ func KeypairListHandler(w http.ResponseWriter, r *http.Request) {
 	formatKeypairsResponse(true, "", "", "", keypairs, w)
 }
 
+func listAllowedKeypairs(authUser datastore.User) ([]datastore.Keypair, error) {
+	switch authUser.Role {
+	case 0:
+		fallthrough
```
rmescandon: You mean the fallthrough?: That's jumping to the next subsequent case.
The default is set outside of the switch but I can change it to be the default switch case instead.
Now, one question for u:
As zero is not a good value to use directly for a role, I need a name for a const instead. Choose one:
- NoAuth
- Invalid
- ...?

And.. its value is good to be 0 or should I take it as -1? (That would involve explicit initialization for all User objects :S)

slimjim777: I think 0 is fine

PR #118, opened 2017-08-09
added validation in user edit. Improve email validation

### LINKED ISSUE #116: User update not validating email or full name
Though email or full name are validated to be not null when adding a new user, that's not happening when editing an existing one

rmescandon:
Fixes #116

### DIFF
--- a/datastore/modeladapter.go
+++ b/datastore/modeladapter.go
@@ -20,7 +20,12 @@
 package datastore
 
 import (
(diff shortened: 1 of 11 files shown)


rmescandon:
- Created adapter for user operations in datastore. That adapter include user params validation before create or update database
- Moved validation steps for creating and updating models to modeladapter.go

PR #149, opened 2017-09-04
switch back to serial-vault name for snap. Bumped version to 2.0-0

rmescandon:
Let's wait for having track 2.0 ready before approving


slimjim777:
Looks like approved too soon.

Did you actually bump the version number to 2.0? That needs to be done in the snapcraft.yaml file and the version setting in the Go code.

PR #152, opened 2017-09-07
support an extended registration protocol with an optional model assertion

pedronis:
This will also cross-check the serial-request with the model assertion if present.

As discussed in

https://forum.snapcraft.io/t/opt-in-accepting-distribution-of-generic-signed-serials/2011

we plan to extend to registration protocol from snapd to include also the model assertion so that it can carry flags to control the serial emission, this is not directly relevant as such so far for the serial-vault itself but the serial-vault needs nevertheless to accept the extra assertion in the stream, and at least lightly cross-check it.

TODO:
* docs updates?

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -123,6 +123,7 @@ func SignHandler(w http.ResponseWriter, r *http.Request) ErrorResponse {
 	dec := asserts.NewDecoder(r.Body)
 	assertion, err := dec.Decode()
 	if err == io.EOF {
+		logMessage("SIGN", "invalid-assertion", "No data supplied for signing")
 		return ErrorEmptyData
 	}
 	if err != nil {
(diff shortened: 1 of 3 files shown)


service/handlers.go
```diff
@@ -117,22 +117,32 @@ func SignHandler(w http.ResponseWriter, r *http.Request) ErrorResponse {
...
-	// Read the full request body
-	data, err := ioutil.ReadAll(r.Body)
+	defer r.Body.Close()
+
+	// Use snapd assertion module to decode the assertions in the request stream
+	dec := asserts.NewDecoder(r.Body)
+	assertion, err := dec.Decode()
+	if err == io.EOF {
+		return ErrorEmptyData
+	}
```
rmescandon: maybe these two error could be evaluated together as io.EOF is also a err != nil. Actually if err == io.EOF nothing is logged using logMessage

service/handlers.go
```diff
@@ -143,6 +153,23 @@ func SignHandler(w http.ResponseWriter, r *http.Request) ErrorResponse {
 		return ErrorInvalidType
 	}
 
+	// Double check the model assertion if present
+	if modelAssert != nil {
+		if modelAssert.Type() != asserts.ModelType {
+			logMessage("SIGN", "invalid-type", "The 2nd assertion type must be 'model'")
```
rmescandon: maybe service/errors.go can be populated with other error messages for this case. A kind of "ErrorInvalidSecondType" or whatever

pedronis:
@rmescandon, trying to answer to the high level questions:

*What should serial vault do with that second model assertion other than checking that the brand-Id and model headers matches in both assertions?

nothing in particular for the moment, the model assertion will always be sent to keep the client behavior uniform,  but the goal of the vault is still to only sign brand-specific serials, the model assertion content (apart basic sanity/validity) is only interesting for the default/fallback registration endpoint in the main store

this at least until we grow anything more that is also controlled by the model

* There should be a model in database related with incoming model assertion?

that is checked by the fact that we check that model matches the serial request and then we look up a model in the db based on the serial-request as before, afaiu, unless I'm misunderstanding the question here;

OTOH as expressed in the TODO longer term it would good to check the signature of the model assertion, that requires though having the relevant keys (usually != from the key to sign serials) or to talk to a service that has all keys (wouldn't work in the factory scenario though)

* Some check related with 'generic-serials' model assertion header?

for  the design the meaning of 'generic-serials: true' is that the model accepts both 'generic'-authority signed and brand signed serials so there is nothing specific to check, models with or without the flag can accept brand-signed serials

service/handlers.go
pedronis: yes, in the changes I dropped by mistake:
logMessage("SIGN", "invalid-assertion", "No data supplied for signing")

it seems putting that back instead of unifying the two cases would be the more backward compatible approach

service/handlers.go
pedronis: ok

pedronis:
@rmescandon I tried to addressed the punctual feedback and add tests for the new error paths, plus a couple more tweaks.

I also tried in the previous comment to address your high-level questions, let me know if you have further doubts or want a more extensive discussion somewhere else.

PR #159, opened 2017-12-14
Generate signing keys

slimjim777:
Adds the ability to generate a named signing-key. This is a time and resource-hungry process, so it is run as a goroutine. This is not a scalable solution as creating too many keys at the same time could cause the server to grind to a halt. The longer term solution would be to use a message queue, adding a separate service to run alongside the serial vault.


datastore/keypairstatusmanager.go
```diff
@@ -0,0 +1,191 @@
...
+CREATE TABLE IF NOT EXISTS keypairstatus (
+	id            serial primary key not null,
+	authority_id  varchar(200) not null,
+	key_name      varchar(200) not null,
+	keypair_id    int references keypair null,
+	status        varchar(20)
+)
+`
+
+const createKeypairStatusSQL = "insert into keypairstatus (authority_id,key_name,status) values ($1,$2,$3) RETURNING id"
```
rmescandon: nitpicking: why reserved words are in upper and lower case in same sql sentence?

datastore/mock_database.go
```diff
@@ -483,6 +489,47 @@ func (mdb *MockDB) ListAccountUsers(authorityID string) ([]User, error) {
 	return mdb.ListUsers()
 }
 
+// CreateKeypairStatusTable mock the database creation
+func (mdb *MockDB) CreateKeypairStatusTable() error {
```
rmescandon: Don't you dare giving a go (pun intended) to mock framework? https://godoc.org/github.com/golang/mock/gomock

service/router.go
```diff
@@ -67,6 +67,10 @@ func AdminRouter() *mux.Router {
 	router.Handle("/v1/keypairs/{id:[0-9]+}/enable", Middleware(http.HandlerFunc(KeypairEnableHandler))).Methods("POST")
 	router.Handle("/v1/keypairs/assertion", Middleware(http.HandlerFunc(KeypairAssertionHandler))).Methods("POST")
 
+	router.Handle("/v1/keypairs/generate", Middleware(http.HandlerFunc(KeypairGenerateHandler))).Methods("POST")
+	router.Handle("/v1/keypairs/status/{authorityID}/{keyName}", Middleware(http.HandlerFunc(KeypairStatusHandler))).Methods("GET")
```
rmescandon: Is it not needed here setting the valid pattern for the identifiers?

settings.yaml
```diff
@@ -34,7 +34,7 @@ urlHost: "serial-vault"
 urlScheme: https
 
 # Enable user authentication using Ubuntu SSO
-enableUserAuth: True
+#enableUserAuth: True
```
rmescandon: commented out on purpose? why?

datastore/keypair.go
```diff
@@ -0,0 +1,138 @@
...
+		log.Println("Error fetching the generated key", err)
+		return "", err
+	}
+
+	// Export the ascii-armored GPG key
+	ks.Status = KeypairStatusExporting
+	if err = Environ.DB.UpdateKeypairStatus(*ks); err != nil {
+		return "", err
+	}
+	out, err := exec.Command("gpg", "--homedir", "~/.snap/gnupg", "--armor", "--export-secret-key", ks.KeyName).Output()
```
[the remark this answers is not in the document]
slimjim777: Not really - it's hardcoded in snapd. There's not a fixed const that I can find in there, but it seems to be hardcoded in a number of places.

datastore/keypairstatusmanager.go
slimjim777: ...you probably don't want to go there! We've been inconsistent with capitals and lowercase in our SQL. Maybe we should stick with uppercase for reserved words as I think that's the standard, but it doesn't make sense to do that in this PR for all SQL.

datastore/mock_database.go
slimjim777: Not right now - there's quite a few other priorities. We have a task to change the tests, so it would be better to handle it there. These tests are an extension of existing ones, so it didn't feel like the right approach to take the newer approach.

service/router.go
slimjim777: These will get validate later. I can include something here, but it means that the rules may not match snapd's validation if that changes.

settings.yaml
slimjim777: Just done for testing. I'll change it back.

PR #160, opened 2017-12-20
Register key with the store

slimjim777:
Register a signing key with the store for a sub-brand. This is the equivalent of a snap register-key.


store/store.go
```diff
@@ -0,0 +1,322 @@
...
+	AuthorityID string `json:"authority-id"`
+	KeyName     string `json:"key-name"`
+}
+
+// RegisterKey registers an account-key assertion with the store
+func RegisterKey(keyAuth KeyRegister, keypair datastore.Keypair) error {
+
+	// Login to the store as the user
+	// Permissions are documented at https://dashboard.snapcraft.io/docs/api/macaroon.html#request-a-macaroon
+	permissions := []string{"edit_account", "modify_account_key"}
```
rmescandon: Any plan to handle these permissions from the UI or are they fix?

store/store.go
```diff
@@ -0,0 +1,322 @@
...
+	_, err = client.Do(r)
+	if err != nil {
+		log.Printf("Error submitting the account-key assertion: %v", err)
+		return err
+	}
+
+	// defer resp.Body.Close()
+
+	// b, err := ioutil.ReadAll(resp.Body)
+	// log.Println(string(b))
```
rmescandon: Why this commented out code?

store/store.go
```diff
@@ -0,0 +1,322 @@
...
+		"account_key_request": string(assert),
+	}
+	d, err := json.Marshal(data)
+	if err != nil {
+		log.Printf("Error marshalling account-key assertion: %v", err)
+		return err
+	}
+
+	// Submit the account-key assertion
+	r, _ := http.NewRequest("POST", storeBaseURL+"account/account-key", bytes.NewReader(d))
```
rmescandon: sounds to me this method to be in a separate place, like a generic http client

store/store.go
```diff
@@ -0,0 +1,322 @@
...
+
+	perm := Permissions{Permissions: permissions}
+	macaroonJSONData, err := json.Marshal(perm)
+	if err != nil {
+		log.Println("Error marshalling the macaroon", err)
+		return "", err
+	}
+
+	// Submit the account-key assertion
+	r, err := http.Post(storeBaseURL+"acl/", "application/json", bytes.NewReader(macaroonJSONData))
```
rmescandon: same as before, maybe it would be nice having all http client operations in same file/place

store/store.go
slimjim777: They are really fixed. These are the minimum that we need for the operations that we are using.

store/store.go
slimjim777: I'll look at it, though the style of the two calls is different.

store/store.go
slimjim777: That's leftover debugging - it was a long painful journey to get to this solution!

store/store.go
slimjim777: I've done it, but it doesn't look great to me. They are different calls: one with custom headers and one without, one with a straight POST and the other with the response being parsed.

PR #161, opened 2018-01-11
Add an API to generate the model assertion

slimjim777:


datastore/mock_database.go
```diff
@@ -195,26 +195,26 @@ func (mdb *MockDB) DeleteAllowedModel(model Model, authorization User) (string,
...
-	model = Model{ID: 7, BrandID: "System", Name: "the-model", KeypairID: 1, AuthorityID: "system", KeyID: "61abf588e52be7a3"}
+	model = Model{ID: 7, BrandID: "system", Name: "the-model", KeypairID: 1, AuthorityID: "system", KeyID: "61abf588e52be7a3"}
 
 	return model, "", nil
 }
 
 // GetKeypair mocks getting a keypair by ID
 func (mdb *MockDB) GetKeypair(keypairID int) (Keypair, error) {
-	keypair := Keypair{ID: 1, AuthorityID: "system", KeyID: "61abf588e52be7a3", Active: true}
+	keypair := Keypair{ID: 1, AuthorityID: "system", KeyID: "UytTqTvREVhx0tSfYC6KkFHmLWllIIZbQ3NsEG7OARrWuaXSRJyey0vjIQkTEvMO", Active: true}
```
rmescandon: These mocks are nice, but I see there is some data that is duplicated in different methods. Maybe you can move it to a common location and take that data from there?

datastore/modelassertmanager.go
```diff
@@ -0,0 +1,154 @@
...
+		gadget           varchar(60) not null,
+		kernel           varchar(60) not null,
+		store            varchar(60),
+		created          timestamp default current_timestamp,
+		modified         timestamp default current_timestamp
+	)
+`
+const createModelAssertSQL = "INSERT INTO modelassertion (model_id,keypair_id,series,architecture,revision,gadget,kernel,store) VALUES ($1,$2,$3,$4,$5,$6,$7,$8) RETURNING id"
+
+const updateModelAssertSQL = `
```
rmescandon: Cosmetic thing: Why some sql is padded and other not?

datastore/modelassertmanager.go
```diff
@@ -0,0 +1,154 @@
...
+	err := db.QueryRow(createModelAssertSQL, m.ModelID, m.KeypairID, m.Series, m.Architecture, m.Revision, m.Gadget, m.Kernel, m.Store).Scan(&createdID)
+	if err != nil {
+		log.Printf("Error creating the model assertion: %v\n", err)
+	}
+	return createdID, err
+}
+
+// UpdateModelAssert updates the model assertion details
+func (db *DB) UpdateModelAssert(m ModelAssertion) error {
+
```
rmescandon: Cosmetic: not blank first line, or at least aligned with other methods if you want to maintain it

datastore/modelassertmanager.go
```diff
@@ -0,0 +1,154 @@
...
+	return err
+
+}
+
+// UpsertModelAssert creates or updates the model assertion headers
+func (db *DB) UpsertModelAssert(m ModelAssertion) error {
+
+	var err error
+
+	if ok := validateModelAssertion(m); ok != nil {
```
rmescandon: This is weird. If ok is not nil, return it... as error?. Why not using same err variable for this?

datastore/modelassertmanager.go
```diff
@@ -0,0 +1,154 @@
...
+	if err := validateNotEmpty("Architecture", m.Architecture); err != nil {
+		return err
+	}
+	if err := validateNotEmpty("Gadget", m.Gadget); err != nil {
+		return err
+	}
+	if err := validateNotEmpty("Kernel", m.Kernel); err != nil {
+		return err
+	}
+	err := validateNotEmpty("Store", m.Store)
```
rmescandon: more cosmetic: I would use same formula as before if err := .. etc and having a return nil at the end, or if you prefer, the opposite:
return validateNotEmpty(...

Current seems a confusing thing of both.

Said this, no matter if left so :)

datastore/modelmanager.go
```diff
@@ -80,7 +80,7 @@ const updateModelForUserSQL = `
 	from account acc
 	inner join useraccountlink ua on ua.account_id=acc.id
 	inner join userinfo u on ua.user_id=u.id
-	where acc.authority_id=m.brand_id and m.id=$1 and u.username=$7`
+	where acc.authority_id=m.brand_id and m.id=$1 and u.username=$8`
```
rmescandon: I don't see the $7

service/handlers_models_test.go
```diff
@@ -29,133 +29,195 @@ import (
...
-	result, _ := sendRequest(t, "GET", "/v1/models/1", nil)
+	result, err := s.parseModelResponse(w)
+	c.Assert(err, check.IsNil)
+	c.Assert(result.Success, check.Equals, true)
+}
 
-	if result.Model.ID != 1 {
-		t.Errorf("Expected model with ID 1, got %d", result.Model.ID)
+func (s *ModelsSuite) TestModelAssertionHeadersHandlerWithoutPermissions(c *check.C) {
+	d := datastore.ModelAssertion{
```
rmescandon: It's fine as is, but just in case, do you know that you can define whatever fields you need into ModelSuite struct and reference them in the different tests?

datastore/modelassertmanager.go
slimjim777: Just the longer ones tend to be padded. The INSERT statements are usually shorter.

datastore/modelassertmanager.go
slimjim777: I think had an error or formatting message about it - I'll recheck.

datastore/modelassertmanager.go
slimjim777: That's how I had it originally, but got a lint error on travis, so had to change it!

PR #162, opened 2018-01-18
The model assertion API call returns multiple assertions

slimjim777:
The model assertion API returns the Account and Account-Key assertions as well as the model assertion, as the device needs these to 'snap ack' the model assertion.

### DIFF
--- a/service/handlers_assertion.go
+++ b/service/handlers_assertion.go
@@ -26,6 +26,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/CanonicalLtd/serial-vault/account"
 	"github.com/CanonicalLtd/serial-vault/datastore"
 	"github.com/snapcore/snapd/asserts"
 )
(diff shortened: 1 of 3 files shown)


service/handlers_assertion.go
```diff
@@ -78,14 +81,33 @@ func ModelAssertionHandler(w http.ResponseWriter, r *http.Request) ErrorResponse
...
+	assertions = append(assertions, signedAssertion)
+
+	// Return successful response with the signed assertions
+	formatAssertionResponse(true, "", "", "", assertions, w)
 	return ErrorResponse{Success: true}
 }
 
+func fetchAssertionFromStore(assertions *[]asserts.Assertion, modelType *asserts.AssertionType, headers []string) error {
+	accountKeyAssertion, err := account.FetchAssertionFromStore(modelType, headers)
+	if err != nil {
```
rmescandon: Why is this method returning an error type if it always takes nil value?

slimjim777: Good point!

PR #164, opened 2018-01-23
Register keypair with the store

slimjim777:
Adding the UI to login to the store and upload a key.


store/store.go
```diff
@@ -126,39 +127,39 @@ func LoginUser(username, password, otp string, permissions []string) (string, st
...
-func submitPOSTRequest(url string, headers map[string]string, data []byte) (*http.Request, error) {
+func submitPOSTRequest(url string, headers map[string]string, data []byte) (*http.Response, error) {
 	r, _ := http.NewRequest("POST", url, bytes.NewReader(data))
 
 	for k, v := range headers {
 		r.Header.Set(k, v)
 	}
 	client := http.Client{}
-	_, err := client.Do(r)
+	resp, err := client.Do(r)
```
rmescandon: why not? 
return client.Do(r)

PR #165, opened 2018-01-31
Configure sub-store mapping for models

slimjim777:
The sub-store mapping allows a model to be pivoted to a sub-store with a new model name


service/handlers_substore.go
```diff
@@ -0,0 +1,189 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2017-2018 Canonical Ltd
```
rmescandon: I wonder if setting only 2018 for new files would be better

service/handlers_substore_test.go
```diff
@@ -0,0 +1,221 @@
...
+var _ = check.Suite(&SubstoreSuite{})
+
+func (s *SubstoreSuite) SetUpTest(c *check.C) {
+	// Mock the database
+	config := config.Settings{KeyStoreType: "filesystem", KeyStorePath: "../keystore", JwtSecret: "SomeTestSecretValue"}
+	datastore.Environ = &datastore.Env{DB: &datastore.MockDB{}, Config: config}
+	datastore.OpenKeyStore(config)
+}
+
+func (s *SubstoreSuite) sendRequest(method, url string, data io.Reader, permissions int, c *check.C) *httptest.ResponseRecorder {
```
rmescandon: Actually you don't need sendRequest, parseSubstoresResponse or parseBooleanResponse to have a pointer receiver to the suite, they are not using anything from the Suite

webapp-admin/src/components/AccountDetail.js
```diff
@@ -0,0 +1,284 @@
+/*
+ * Copyright (C) 2017-2018 Canonical Ltd
```
rmescandon: 2018?

webapp-admin/src/components/SubstoreForm.js
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (C) 2017-2018 Canonical Ltd
```
rmescandon: 2018

service/handlers_substore.go
slimjim777: Okay, I can do that

PR #169, opened 2018-02-13
Add the optional required-snaps model assertion header

slimjim777:


datastore/modelassertmanager.go
```diff
@@ -36,40 +36,42 @@ const createModelAssertTableSQL = `
...
+	ID            int       `json:"id"`
+	ModelID       int       `json:"model_id"`
+	KeypairID     int       `json:"keypair_id"`
+	Series        int       `json:"series"`
+	Architecture  string    `json:"architecture"`
+	Revision      int       `json:"revision"`
+	Gadget        string    `json:"gadget"`
+	Kernel        string    `json:"kernel"`
+	Store         string    `json:"store"`
+	RequiredSnaps string    `json:"required-snaps"`
```
rmescandon: won't that dash cause problems in json?. You'd better use 'required_snaps' i think

slimjim777: The js code actually handles that in this case. I've wanted to switch to underscores over time, but this change was needed quickly.

PR #170, opened 2018-02-15
Add an API key field to the user

slimjim777:
Add the user API key to the UI


slimjim777:
This is a change of approach to what we discussed as we many need multiple API keys per account, so we can disable access to the API for one user without affecting other users. Eventually, we may need additional API roles e.g. one that can just get signing-logs, another that can create models.

datastore/modeladapter.go
```diff
@@ -194,21 +196,17 @@ func validateKeypairIDUser(keypairIDUser int) error {
...
+		return apiKey, nil
 	}
 
 	apiKey, err := generateAPIKey()
-	if err != nil {
-		return err
-	}
-	model.APIKey = apiKey
-	return nil
+	return apiKey, err
```
rmescandon: return generateAPIKey() I think

PR #171, opened 2018-02-16
Admin API call for signing logs

slimjim777:
This refactors the signing log list handler to a separate package so it can be called from the UI or the Admin API. This just focuses on the one handler and creates a bit of duplication in the refactoring to minimise the number of changes. The idea is to make this simpler to review and creating less noise.

The idea is to prototype the approach that can be used for the other handlers. This separates out the fetching of the user from the check to see if the role has access to the function - which may help in the future with a macaroon-style approach in the future.

### DIFF
--- a/datastore/usermanager.go
+++ b/datastore/usermanager.go
@@ -46,6 +46,7 @@ const createAccountUserLinkTableSQL = `
 const listUsersSQL = "select id, username, name, email, userrole, api_key from userinfo order by username"
 const getUserSQL = "select id, username, name, email, userrole, api_key from userinfo where id=$1"
 const getUserByUsernameSQL = "select id, username, name, email, userrole, api_key from userinfo where username=$1"
+const getUserByAPIKeySQL = "select id, username, name, email, userrole, api_key from userinfo where api_key=$1 and username=$2"
 const findUsersSQL = "select id, username, name, email, userrole, api_key from userinfo where username like '%$1%' or name like '%$1%'"
 const createUserSQL = "insert into userinfo (username, name, email, userrole, api_key) values ($1,$2,$3,$4,$5) RETURNING id"
 const updateUserSQL = "update userinfo set username=$1, name=$2, email=$3, userrole=$4, api_key=$6 where id=$5"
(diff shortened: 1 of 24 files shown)


datastore/usermanager.go
```diff
@@ -217,6 +218,16 @@ func (db *DB) GetUserByUsername(username string) (User, error) {
 	return user, err
 }
 
+// GetUserByAPIKey fetches a single user from database
+func (db *DB) GetUserByAPIKey(apiKey, username string) (User, error) {
```
rmescandon: I guess GetUserByAPIKeyAndUsername is a bit verbose, isn't it?

service/auth.go
```diff
@@ -34,7 +35,7 @@ func checkPermissionsAndGetUserFromJWT(w http.ResponseWriter, r *http.Request, m
 }
 
 func getUserFromJWT(w http.ResponseWriter, r *http.Request) (datastore.User, error) {
-	token, err := JWTCheck(w, r)
+	token, err := utils.JWTCheck(w, r)
```
rmescandon: Go doesn't recommend having a utils package. What about something like 'auth', 'sso' or even 'jwt'?. Actually, I think Go talibans would be happy with something like jwt.Check(w,r)

service/signinglog/handlers_adminapi.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"net/http"
+
+	"github.com/CanonicalLtd/serial-vault/datastore"
+	"github.com/CanonicalLtd/serial-vault/service/utils"
+)
+
+// APIListHandler is the API method to fetch the log records from signing
+func APIListHandler(w http.ResponseWriter, r *http.Request) {
```
rmescandon: probably later, this requires another specific subpackage like signinglog.api or similar?

service/utils/utils.go
```diff
@@ -0,0 +1,47 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
```
rmescandon: Definitively that utils package deserves another name, though I don't know which one, tbh

datastore/usermanager.go
slimjim777: Yes, that's what I was thinking

PR #172, opened 2018-02-19
Admin API for keypair list

slimjim777:
This is adds another list API and helps to identify the pattern that will be needed for the other APIs


service/keypair/actions.go
```diff
@@ -0,0 +1,71 @@
...
+	"net/http"
+
+	"github.com/CanonicalLtd/serial-vault/datastore"
+	"github.com/CanonicalLtd/serial-vault/service/auth"
+	"github.com/CanonicalLtd/serial-vault/service/response"
+)
+
+// ListResponse is the JSON response from the API Keypairs method
+type ListResponse struct {
+	Success      bool                `json:"success"`
```
rmescandon: not now, but maybe in future you could use a composition of structs, having the sucess and the error things in a common struct for any response

slimjim777: Yep

PR #175, opened 2018-02-24
Integrate govendor

rmescandon:
Updated run-checks with more checks and skipping vendor or not relevant folders
Fix code as run-checks demands
Update .md files lists to fit mdlint tests
Tested deb package successfully built in nightly ppa.
  - deb build tests by using `go tests` but cannot execute run-checks as it uses tools that should be downloaded in build time, what is not possible within ppa. Run-checks however are executed in github phase, so there's any trouble on skip them when building.

Note: Commented out shellcheck in run-checks script because there are a lot of warnings that has large correction. Please do this in another PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ stage
 *.tar.bz2
 snap/.snapcraft
 .tests-extras
+vendor/*/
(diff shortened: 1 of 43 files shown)


rmescandon:
Probably a `govendor sync` makes more sense than using go get or godeps there

PR #177, opened 2018-03-01
Admin API for sub-store models

slimjim777:


datastore/usermanager.go
```diff
@@ -220,6 +220,10 @@ func (db *DB) GetUserByUsername(username string) (User, error) {
 
 // GetUserByAPIKey fetches a single user from database
 func (db *DB) GetUserByAPIKey(apiKey, username string) (User, error) {
+	if len(apiKey) == 0 || len(username) == 0 {
```
rmescandon: Worths having these two things in a different checks or that could be a security breach?

service/substore/handlers_adminapi.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/gorilla/mux"
+)
+
+// APIListHandler is the API method to fetch the sub-store models
+func APIListHandler(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+
+	// Get the user and API key from the header
+	username := r.Header.Get("user")
+	apiKey := r.Header.Get("api-key")
```
rmescandon: Maybe there could be a helper method with this code that is repeated across these different handlers:

    // Get the user and API key from the header
	username := r.Header.Get("user")
	apiKey := r.Header.Get("api-key")

	// Find the user by API key
	user, err := datastore.Environ.DB.GetUserByAPIKey(apiKey, username)
	if err != nil {
		response.FormatStandardResponse(false, "error-auth", "", err.Error(), w)
		return
	}

datastore/usermanager.go
slimjim777: In two different check, if both user and api-key are omitted then it will fail with the first check. The  you correct that, and it fails with the second check. I'm not sure that it helps from a user standpoint

PR #181, opened 2018-03-13
Add a health check API to serial-vault

ejfinneran:
`/v1/health` checks to ensure that the database is accessible using the
db.Ping() function.

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -94,8 +94,10 @@ func HealthHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
 	err := datastore.Environ.DB.HealthCheck()
 	var database string
+
 	if err != nil {
 		database = err.Error()
+		w.WriteHeader(http.StatusBadRequest)
 	} else {
 		database = "healthy"
 	}
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -89,8 +89,8 @@ func VersionHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-// HealthHander is the API method to return if the app is up and db.Ping() doesn't return an error
-func HealthHander(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


service/handlers.go
```diff
@@ -84,6 +89,23 @@ func VersionHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// HealthHander is the API method to return if the app is up and db.Ping() doesn't return an error
+func HealthHander(w http.ResponseWriter, r *http.Request) {
```
slimjim777: Minor comment: Hander should be Handler
However, we are in the process of a refactor and removing the team Handler altogether from these functions.

service/handlers_test.go
```diff
@@ -643,6 +643,45 @@ func TestVersionHandler(t *testing.T) {
...
+
+}
+
+func TestHealthHandlerUnhealthy(t *testing.T) {
+
+	config := config.Settings{KeyStoreType: "filesystem", KeyStorePath: "../keystore", JwtSecret: "SomeTestSecretValue"}
+	datastore.Environ = &datastore.Env{DB: &datastore.ErrorMockDB{}, Config: config}
+	response, _ := sendRequestHealth(t, "GET", "/v1/health", nil)
+
+	if response.Code != 500 {
```
slimjim777: This test is failing as the method always returns an HTTP 200 response. You'll need to set the response in the HealthCheck method. Also, we prefer HTTP 400 errors for understood application errors - that way we know that if we get an error 500 we know that something unexpected has happened.

So, for example, the health check may actually fail when the service is first deployed as the postgresql unit takes longer to start up and initialize than the SV units.

ejfinneran:
I fixed the tests and function names but it looks like `db.Ping()` doesn't handle connections dying very well so it's sending back false positives right now.

Need to look at better ways check DB health before merging this.

ejfinneran:
I opted for running a simple query (`select 1`) . I tested the endpoint while terminating the database container and the endpoint started return 400 as expected.

PR #184, opened 2018-03-16
Refactor API handlers

slimjim777:
The final episode of the refactoring of the API handlers.
A few more tests added in the process.


service/app/handlers_app_user.go
```diff
@@ -35,11 +35,14 @@ import (
 	"github.com/CanonicalLtd/serial-vault/crypt"
 	"github.com/CanonicalLtd/serial-vault/datastore"
 	"github.com/CanonicalLtd/serial-vault/random"
+	svlog "github.com/CanonicalLtd/serial-vault/service/log"
```
rmescandon: is this prefix needed?. I don't see any other log package here.

service/log/log.go
```diff
@@ -0,0 +1,28 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2018-2019 Canonical Ltd
```
rmescandon: 2019 is the year of Serial Vault, i see it

service/log/log.go
```diff
@@ -0,0 +1,28 @@
...
+ *
+ */
+
+package log
+
+import "log"
+
+// Message logs a message in a fixed format so it can be analyzed by log handlers
+// e.g. "METHOD CODE descriptive reason"
+func Message(method, code, reason string) {
```
rmescandon: Maybe not now but I would be good implementing a whole logging system, with log level to be configured in settings.
I tried go-logging and works really well
```
package log

import (
	"os"

	logging "github.com/op/go-logging"
)

var l = logging.MustGetLogger("serialvault")

// InitLogger initializes logger for backend with the specified level
func InitLogger(level logging.Level) {
	backend := logging.NewLogBackend(os.Stderr, "", 0)

	format := logging.MustStringFormatter(
		`%{color}%{time:15:04:05.000} %{shortfunc} ▶ %{level:.4s} %{id:03x}%{color:reset} %{message}`,
	)
	backendFormatter := logging.NewBackendFormatter(backend, format)

	backendLeveled := logging.AddModuleLevel(backendFormatter)
	backendLeveled.SetLevel(level, "")

	logging.SetBackend(backendLeveled)
}

// Errorf calls logger in eror level with format
func Errorf(format string, args ...interface{}) {
	l.Errorf(format, args)
}

// Error calls logger in error level
func Error(args ...interface{}) {
	l.Error(args)
}

// Warningf calls logger in warning level with format
func Warningf(format string, args ...interface{}) {
	l.Warningf(format, args)
}

// Warning calls logger in warning level
func Warning(args ...interface{}) {
	l.Warning(args)
}
...
```

service/app/handlers_app_user.go
slimjim777: Doesn't look like it. I must have replaced all the original log.Printf commands.

slimjim777: Actually it is used, but hidden in this diff (line 25)

rmescandon: ok, thanks

PR #185, opened 2018-03-20
Remove system user service

slimjim777:
Also bumping the version.


manage/client.go
```diff
@@ -36,6 +36,8 @@ import (
 	"github.com/snapcore/snapd/asserts"
 )
 
+var deviceKey = "./keystore/TestDeviceKey.asc"
```
rmescandon: var or const?

slimjim777: It's var as it is overwritten in the test. We used the same trick for IndexTemplate in handlers_app.go.

PR #186, opened 2018-03-21
Add a new SyncUser role.

slimjim777:
Add sync API access to accounts


datastore/usermanager.go
```diff
@@ -86,20 +86,22 @@ const alterUserAPIKeyNotNullable = `alter table userinfo
...
 // * Standard:	role for regular users. This is the less privileged role
+// * SyncUser:	role for users that will used the Sync API
 // * Admin:		role for admin users, including standard role permissions but not superuser ones
 // * Superuser:	role for users having all the permissions
 const (
 	Invalid   = iota       // 0
 	Standard  = 100 * iota // 100
 	Admin                  // 200
 	Superuser              // 300
+	SyncUser  = 150
```
rmescandon: mm.. what about having direct assignments instead of using iota?. Looks weird now having a mix of iota, value or nothing.

Another option, I've seen in lxd project that they redefine the int type for situations like this:

```
type operationClass int

const (
	operationClassTask      operationClass = 1
	operationClassWebsocket operationClass = 2
	operationClassToken     operationClass = 3
)

func (t operationClass) String() string {
	return map[operationClass]string{
		operationClassTask:      "task",
		operationClassWebsocket: "websocket",
		operationClassToken:     "token",
	}[t]
}
```

service/user/actions.go
```diff
@@ -134,7 +136,7 @@ func formatUserResponse(user datastore.User, w http.ResponseWriter) error {
 
 	// Encode the response as JSON
 	if err := json.NewEncoder(w).Encode(response); err != nil {
-		log.Println("Error forming the user response.")
+		svlog.Error("error-user-response", err)
```
rmescandon: I guess the idea behind this is searching for the whole message under 'error-user-response' and printing it along with the tag, right?

slimjim777: It's mainly for log analytics. It makes it simpler to do reporting on a specific type of error based on the code.

rmescandon: I wonder if having an int error_code and a free error_message text could be a solution. Analytics is a M2M interface that maybe prefers an integer (just a thought)

PR #187, opened 2018-03-22
Fix the sub-store validation to match snapd

slimjim777:


datastore/modeladapter_test.go
```diff
@@ -43,17 +42,6 @@ func TestModelNameEmpty(t *testing.T) {
 	}
 }
 
-func TestModelNameInvalidChar(t *testing.T) {
```
rmescandon: This kind of changes deserve some tests

slimjim777: This test is no longer valid... that's why I removed it.

rmescandon: Sure, but I was pointing if another similar test would be needed like TestValidateCaseInsensitive()

slimjim777: As discussed, this is covered in TestModelNameUpperChar

PR #192, opened 2018-03-29
Factory sync task

slimjim777:


cmd/factory-sync/main.go
```diff
@@ -0,0 +1,58 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
```
rmescandon: why not directly "sync" command?

datastore/accountadapter.go
```diff
@@ -84,3 +84,13 @@ func (db *DB) UpdateAccount(account Account, authorization User) error {
 		return nil
 	}
 }
+
+// SyncAccount validates permissions and stores an account in the database
+func (db *DB) SyncAccount(account Account) (string, error) {
+	err := validateAuthorityID(account.AuthorityID)
+	if err != nil {
+		return "error-validate-account", err
```
rmescandon: I would wrap both message and error in a single thing

datastore/accountmanager.go
```diff
@@ -84,6 +84,17 @@ const listNotUserAccountsSQL = `
...
 
+const syncUpsertAccountSQL = `
+	WITH upsert AS (
+		update account set authority_id=$2, assertion=$3, resellerapi=$4
+		where id=$1
+		RETURNING *
+	)
+	insert into account (id,authority_id,assertion,resellerapi)
+	select $1, $2, $3, $4
+	where not exists (select * from upsert)
```
rmescandon: Have you tried on conflict do update sentence. I think it reduces a bit the code

datastore/accountmanager.go
```diff
@@ -214,6 +225,17 @@ func (db *DB) putAccount(account Account) (string, error) {
...
 
+// syncAccount stores an account in the database
+func (db *DB) syncAccount(account Account) (string, error) {
+	_, err := db.Exec(syncUpsertAccountSQL, account.ID, account.AuthorityID, account.Assertion, account.ResellerAPI)
+	if err != nil {
+		log.Printf("Error updating the database account: %v\n", err)
+		return "", err
+	}
+
+	return "", nil
```
rmescandon: not sure what that first string returning value is, but it is always ""

sync/cloud.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+// SendRequest sends the request to the serial vault
+var SendRequest = func(method, url, endpoint, username, apikey string, data []byte) (*http.Response, error) {
+	log.Infof("Call the cloud %s", url+endpoint)
+	r, _ := http.NewRequest(method, url+endpoint, bytes.NewReader(data))
+	r.Header.Set("user", username)
+	r.Header.Set("api-key", apikey)
+
+	client := http.Client{}
```
rmescandon: Maybe the client could be in a var instead of creating it every time

sync/start.go
```diff
@@ -0,0 +1,78 @@
...
+	"errors"
+
+	"github.com/CanonicalLtd/serial-vault/service/log"
+)
+
+// StartCommand starts the sync process
+type StartCommand struct {
+	URL      string `short:"s" long:"svurl" description:"Sync URL for the cloud serial-vault" default:"https://serial-vault-partners.canonical.com/api/"`
+	Username string `short:"u" long:"user" description:"Sync username for the cloud serial-vault"`
+	APIKey   string `short:"a" long:"apikey" description:"Sync API key for the cloud serial-vault"`
```
rmescandon: Are some of these parameters mandatory. In such case I would prefer using the tool with positional args:
```
        Args struct {
		Username      string `positional-arg-name:"username" required="yes"`
		APIKey string `positional-arg-name:"api-key" required="yes"`
		URL         string `positional-arg-name:"url" optional="true"`
	} `positional-args:"yes"`
```

sync/start.go
```diff
@@ -0,0 +1,78 @@
...
+// Execute the sync for the factory
+func (cmd StartCommand) Execute(args []string) error {
+	withErrors := false
+
+	if len(cmd.URL) == 0 || len(cmd.Username) == 0 || len(cmd.APIKey) == 0 {
+		return errors.New("The cloud serial vault URL, username and API key must be provided")
+	}
+
+	// Open the connection to the factory database
+	openDatabase()
```
rmescandon: What if database cannot be open. Does this control the error?

sync/sync.go
```diff
@@ -0,0 +1,48 @@
...
+	// Check that the database has not been set e.g. by a mock
+	if datastore.Environ.DB != nil {
+		return
+	}
+
+	log.Infof("Open the settings file: %s", Sync.SettingsFile)
+	config.ReadConfig(&datastore.Environ.Config, Sync.SettingsFile)
+
+	// Open the connection to the database
+	datastore.OpenSysDatabase(datastore.Environ.Config.Driver, datastore.Environ.Config.DataSource)
```
rmescandon: This method looks like too ideal... no error check?

datastore/accountmanager.go
slimjim777: I think that's only available at later postgresql versions, so I've avoided it deliberately... until the world moves on.

sync/start.go
slimjim777: They are not - these may be read in from the config file. I'm still considering moving these to the config file and I need to add a daemon mode for the snap.

sync/start.go
slimjim777: We get an early exit if the database is not open - so it works as expected

sync/sync.go
slimjim777: No error check that is caught - the command exists with an error, which is correct

cmd/factory-sync/main.go
slimjim777: Just to make it clearer that it is only for deployment in a factory. In some ways this could be seen as a separate project.

PR #196, opened 2018-04-06
Sqlite in the factory snap

slimjim777:
Use sqlite3 in the factory instead of postgresql. This is more lightweight and circumvents the issues that postgresql has with running as root.


datastore/database_pg.go
```diff
@@ -0,0 +1,46 @@
...
+	"log"
+
+	_ "github.com/lib/pq" // postgresql driver
+)
+
+// openPostgreSQLDatabase return an open database connection for an postgresql database
+func openPostgreSQLDatabase(driver, dataSource string) {
+	// Open the database connection
+	db, err := sql.Open(driver, dataSource)
+	if err != nil {
```
rmescandon: There is a common code in both drivers opening. I wonder if that could be in a common file/site

datastore/noncemanager.go
```diff
@@ -87,7 +89,20 @@ func (db *DB) CreateDeviceNonce() (DeviceNonce, error) {
 	}
 
 	// Create the nonce in the database
-	_, err = db.Exec(createDeviceNonceSQL, nonce.Nonce, nonce.TimeStamp)
+	if Environ.Config.Driver == "sqlite3" {
+		// Need to generate our own ID
+		var nextID int
+		err = db.QueryRow(maxIDDeviceNonceSQLite).Scan(&nextID)
```
rmescandon: not sure if this snippet taking the last inserted id is of help in this case:
```

// CreateQueueResource creates a new application in the datastore
func (d *DS) CreateQueueResource(res *QueueResource) (int, error) {
	result, err := d.Exec(insertQueueResourceSQL, res.Status, res.Location, res.Resources, res.Due.UTC())
	if err != nil {
		return -1, err
	}

	id, err := result.LastInsertId()
	return int(id), err
}
```

datastore/database_pg.go
slimjim777: I did wonder that, but thought it would be safer to have the repeat, since there is some magic happening with the `_` import

datastore/noncemanager.go
slimjim777: This is a different use-case. With the serial-vault we have an auto-generated `id` in postgresql, but sqlite needs us to supply a value for the `id` field. What lastInsertId does is returns a hidden `rowid` field from the record... not quite the same thing

PR #198, opened 2018-04-09
Sync signing logs

slimjim777:
Add sync API for signing logs and ability to sync the logs from the factory


sync/client.go
```diff
@@ -152,6 +152,33 @@ func (c *FactoryClient) Models() error {
...
+	// Send each signing log to the cloud
+	for _, l := range logs {
+		success, err := SendSigningLog(c.URL, c.Username, c.APIKey, l)
+		if err != nil || !success {
+			// Leave this one till the next sync
+			continue
+		}
+
+		// Mark the sync as done
+		err = datastore.Environ.DB.SyncUpdateSigningLog(l.ID)
```
bergotorino: would there be a gain if implemented via Tx type

slimjim777: Not in this case. This is 2 different databases being updated. For each record, the cloud SV is updated and then the local db is updated for that record. If the internet connection in the factory goes down, you wouldn't want to lose the fact that you'd successfully synced some records - that would lead to them being resent later.

PR #200, opened 2018-04-11
Form to cache test logs

slimjim777:
Add a form to the factory signing service that allows test logs to be cached to disk

### DIFF
--- a/service/router.go
+++ b/service/router.go
@@ -36,6 +36,7 @@ import (
 	"github.com/CanonicalLtd/serial-vault/service/signinglog"
 	"github.com/CanonicalLtd/serial-vault/service/store"
 	"github.com/CanonicalLtd/serial-vault/service/substore"
+	"github.com/CanonicalLtd/serial-vault/service/testlog"
 	"github.com/CanonicalLtd/serial-vault/service/user"
 	"github.com/CanonicalLtd/serial-vault/usso"
 	"github.com/gorilla/mux"
@@ -55,6 +56,12 @@ func SigningRouter() *mux.Router {
 	router.Handle("/v1/pivot", Middleware(ErrorHandler(pivot.Model))).Methods("POST")
 	router.Handle("/v1/pivotmodel", Middleware(ErrorHandler(pivot.ModelAssertion))).Methods("POST")
 	router.Handle("/v1/pivotserial", Middleware(ErrorHandler(pivot.SerialAssertion))).Methods("POST")
+
+	// Test log upload routes (only in the factory)
+	if datastore.Environ.Config.Driver == "sqlite3" {
+		router.Handle("/testlog", Middleware(http.HandlerFunc(testlog.Index))).Methods("GET")
+		router.Handle("/testlog", Middleware(http.HandlerFunc(testlog.Submit))).Methods("POST")
+	}
 
 	return router
 }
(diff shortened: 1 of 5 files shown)


service/testlog/handlers_testlog.go
```diff
@@ -0,0 +1,123 @@
...
+		return
+	}
+
+	// Set up the path for the file
+	path := os.Getenv(paramsEnvVar)
+	if len(path) == 0 {
+		path = "."
+	}
+	path = path + "/files"
+	_ = os.Mkdir(path, 0755)
```
bergotorino: unlikely but still shouldn't it be logged when mkdir fails - like config variable points to a not writable location

service/router.go
```diff
@@ -56,6 +57,12 @@ func SigningRouter() *mux.Router {
 	router.Handle("/v1/pivotmodel", Middleware(ErrorHandler(pivot.ModelAssertion))).Methods("POST")
 	router.Handle("/v1/pivotserial", Middleware(ErrorHandler(pivot.SerialAssertion))).Methods("POST")
 
+	// Test log upload routes (only in the factory)
+	if datastore.Environ.Config.Driver == "sqlite3" {
```
bergotorino: Does it mean that the mode cloud vs factory is recognized based on the database driver being used?

service/testlog/handlers_testlog.go
```diff
@@ -0,0 +1,123 @@
...
+		formatUploadResponse(w, http.StatusBadRequest, err.Error())
+		return
+	}
+
+	// Set up the path for the file
+	path := os.Getenv(paramsEnvVar)
+	if len(path) == 0 {
+		path = "."
+	}
+	path = path + "/files"
```
rmescandon: use filepath.Join(). Who knows what format has the received path

service/testlog/handlers_testlog.go
slimjim777: It will fail if the directory already exists as well. This will only be used within a snap, so I think its fine for now. I could do a check if directory-exists-and-create-it-if-not, but I think this is a little for efficient.

service/router.go
slimjim777: Yes, not ideal, but the only way that I can see to identify if we're in the factory. The form doesn't have any CSRF protection, to make it simpler to use from the device test tools, so it's not something that we should deploy in the cloud.

service/testlog/handlers_testlog.go
slimjim777: Nice - another secret Go module

service/testlog/handlers_testlog.go
```diff
@@ -0,0 +1,130 @@
...
+	path := os.Getenv(paramsEnvVar)
+	if len(path) == 0 {
+		path = "."
+	}
+	path = filepath.Join(path, "/files")
+
+	// Attempt to create the path, if it doesn't exist
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		if err = os.Mkdir(path, 0755); err != nil {
+			formatUploadResponse(w, http.StatusBadRequest, err.Error())
```
bergotorino: shouldn't it exit when creating the dir fails?

slimjim777: Good catch!

PR #201, opened 2018-04-12
Sync API for test logs

slimjim777:
Adding the sync API for testlogs:
 - use testlog record in factory instead of the filesystem
 - add the brand/model to the web form (so permissions can be determined in the cloud)
 - add testlog (factory and cloud)
 - list testlog (cloud)

### DIFF
--- a/datastore/testlogmanager.go
+++ b/datastore/testlogmanager.go
@@ -24,7 +24,6 @@ import (
 	"database/sql"
 	"errors"
 	"log"
-	"strings"
 	"time"
 )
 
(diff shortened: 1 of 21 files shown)


datastore/testlogmanager.go
```diff
@@ -0,0 +1,144 @@
...
+	Filename string    `json:"filename"`
+	Data     string    `json:"data"`
+	Created  time.Time `json:"created"`
+	Synced   time.Time `json:"synced"` // used to indicate it has been synced to an external system
+}
+
+// CreateTestLogTable creates the database table for a test log
+func (db *DB) CreateTestLogTable() error {
+	_, err := db.Exec(createTestLogTableSQL)
+	if err != nil {
```
rmescandon: return err

datastore/testlogmanager.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return nil
+}
+
+// CreateTestLog keeps a record of a test log
+func (db *DB) CreateTestLog(testLog TestLog) error {
+	var err error
+	// Validate the data
+	if strings.TrimSpace(testLog.Brand) == "" || strings.TrimSpace(testLog.Model) == "" || strings.TrimSpace(testLog.Filename) == "" || strings.TrimSpace(testLog.Data) == "" {
```
rmescandon: I wonder if a generic method to check and array of strings not to be empty... like notEmpty(args... string)

datastore/testlogmanager.go
```diff
@@ -0,0 +1,144 @@
...
+		return errors.New("The brand, model, filename and file (base64-encoded) must be supplied")
+	}
+
+	// Create the signing log in the database
+	if Environ.Config.Driver == "sqlite3" {
```
rmescandon: Uf.. this thing of having to deal with different drivers.. not in this pr, but in another maybe you find a kind of decorator  to have this thing for every needed query in a common place (not sure if that is possible either)

slimjim777: I've toyed with the idea of putting the factory queries onto a separate interface, but the issue is that the factory uses the "standard" signing service. I'm not sure how to improve this at this stage. The database access has always felt a little messy.

PR #206, opened 2018-04-24
Admin API to validate a serial assertion

slimjim777:


datastore/substoremanager.go
```diff
@@ -152,6 +151,28 @@ func (db *DB) GetSubstore(fromModelID int, serialNumber string) (Substore, error
 	return store, nil
 }
 
+// GetSubstoreModel fetches a sub-store in the database
+func (db *DB) GetSubstoreModel(brand, model, serialNumber string) (Substore, error) {
+	store := Substore{}
+
+	var row *sql.Row
```
bergotorino: this line could be omitted in favor of := assignment below

service/assertion/actions_check.go
```diff
@@ -0,0 +1,67 @@
...
+
+	err := auth.CheckUserPermissions(authUser, datastore.Admin, apiCall)
+	if err != nil {
+		response.FormatStandardResponse(false, response.ErrorAuth.Code, "", "", w)
+		return
+	}
+
+	// Check that the account is accessbible by the user
+	if _, err = datastore.Environ.DB.GetAllowedAccount(assertion.HeaderString("brand-id"), authUser); err != nil {
+		response.FormatStandardResponse(false, "invalid-account", "", "", w)
```
bergotorino: in this function.. mixing error codes as enums with literals.

Issue #144, opened 2017-08-17
Wishlist: Hide brand/authority if there is only one available


rmescandon:
In case there is only one brand/authority it would be nice having that in a general place visible from everywhere in the portal instead of showing every model, key or assertion linked to that brand/authority. That way, the admin accessing the UI would know that any model, key or assertion is related to same brand/authority

PR #211, opened 2018-05-15
Updated user interface with account selection

slimjim777:
Each page is filtered by a selected account in the menu.


slimjim777:
Closes #144

PR #221, opened 2018-08-17
Add new model assertion fields

slimjim777:
Add support for the model assertion fields recently introduced for uc18


slimjim777:
Closes #221

service/assertion/actions.go
```diff
@@ -102,16 +102,42 @@ func CreateModelAssertionHeaders(m datastore.Model) (map[string]interface{}, dat
...
+	if len(assert.Classic) != 0 {
+		headers["classic"] = assert.Classic
+	}
+
+	if len(assert.DisplayName) != 0 {
+		headers["display-name"] = assert.DisplayName
+	}
+
+	// Some headers are required for Ubuntu Core, whilst optional or invalid for Classic
+	if headers["classic"] == "true" {
```
rmescandon: Is validated somehow the case that the model creator wanted to set this as true but wrote "True" or similar?. It would be bad that it goes to false because of a misspelling.
If not, you could use a switch here to evaluate "true", "false" and default

datastore/modelassertmanager.go
```diff
@@ -37,23 +37,26 @@ const createModelAssertTableSQL = `
 		kernel           varchar(60) not null,
 		store            varchar(60),
 		required_snaps   text default '',
+		base             varchar(20) default '',
```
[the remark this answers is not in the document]
slimjim777: That's what started all this :) It's a new field for uc18:

https://forum.snapcraft.io/t/model-assertions-for-core18/6870

service/assertion/actions.go
slimjim777: It's done in the UI - attempts to correct the typos. But it could be called from the API, so I'll add an additional check here.

PR #223, opened 2018-08-29
Validate the model does not already exist

### LINKED ISSUE #222: Check that the brand/model name does not exist when updating it
Currently, creating a model checks that the name does not already exist. But when updating a model it is possible to create a duplicate.

slimjim777:
Closes #222

### DIFF
--- a/datastore/modeladapter.go
+++ b/datastore/modeladapter.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 
 /*
- * Copyright (C) 2016-2017 Canonical Ltd
+ * Copyright (C) 2016-2018 Canonical Ltd
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


datastore/modeladapter.go
```diff
@@ -79,6 +79,20 @@ func (db *DB) UpdateAllowedModel(model Model, authorization User) (string, error
 	}
 	model.APIKey = apiKey
 
+	// Get the existing model using the ID
+	m, err := db.getModel(model.ID)
+	if err != nil {
```
rmescandon: Could it happen other kind of error? Maybe the check should be more explicit here and check that the db returned error is exactly that the model is not found

slimjim777: I did consider that, but the error message would be a fairly raw database error. I thought that the sanitized message would be better.

rmescandon: At least you should log that raw db error, just in case it happens, for not to drive crazy searching what has happened

PR #228, opened 2018-12-13
Updated model and substore create API to return created object.

matiasb:
Updated 'create' endpoints to return the created instance details.

Return created model and substore instances on a successful POST to the respective endpoints.

Model create: https://pastebin.canonical.com/p/48gzpyxcNP/
Substore create: https://pastebin.canonical.com/p/mxw6myR43Z/


pedronis:
I see, the permission check is done beforehand, but then is still unclear why not return an internal error instead of a zero instance

PR #229, opened 2018-12-13
Added admin endpoint to GET a substore by modelID and serial.

matiasb:
Reproduces auth logic from other endpoints, checking user role and access.
Endpoint only enabled for the admin API.

Built on top of #228.

Example: https://pastebin.canonical.com/p/VKzCVp2hmR/

### DIFF
--- a/datastore/substoreadapter.go
+++ b/datastore/substoreadapter.go
@@ -41,6 +41,20 @@ func (db *DB) ListSubstores(accountID int, authorization User) ([]Substore, erro
 	}
 }
 
+// GetAllowedSubstore return the sub-store if the user is authorized to see it
+func (db *DB) GetAllowedSubstore(modelID int, serial string, authorization User) (Substore, error) {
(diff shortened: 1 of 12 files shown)


datastore/substoreadapter.go
```diff
@@ -41,6 +41,20 @@ func (db *DB) ListSubstores(accountID int, authorization User) ([]Substore, erro
...
+func (db *DB) GetAllowedSubstore(modelID int, serial string, authorization User) (Substore, error) {
+	switch authorization.Role {
+	case Invalid: // Authentication disabled
+		fallthrough
+	case Superuser:
+		return db.GetSubstore(modelID, serial)
+	case Admin:
+		return db.GetSubstoreFilteredByUser(modelID, serial, authorization.Username)
+	default:
+		return Substore{}, nil
```
pedronis: again, my impulse here would be to return an internal error instead, not sure how diffuse this style is, but is peculiar

PR #231, opened 2019-04-16
add detail to errors where useful and log and return them in response…

PreservedKillick:
…s for models and substores

### DIFF
--- a/datastore/modelassertmanager.go
+++ b/datastore/modelassertmanager.go
@@ -20,8 +20,7 @@
 package datastore
 
 import (
-	"errors"
-	"log"
+	"fmt"
 	"time"
 )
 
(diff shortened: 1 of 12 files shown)


datastore/modelassertmanager.go
```diff
@@ -155,41 +154,46 @@ func (db *DB) deleteModelAssert(modelID int) error {
 
 	_, err = db.Exec(deleteModelAssertSQL, modelID)
 	if err != nil {
-		log.Printf("Error deleting the model assertion: %v\n", err)
```
matiasb: Any reason to drop the logging here? Given the raised error is not always caught/displayed later, this maybe the only information we keep in this case (similar for other cases below too).

PreservedKillick: Ah you're right, one of the callers is not handling this error at all.  I fixed that and went through the other funcs below where I've eliminated logging of errors that are getting thrown to make sure that callers were also either logging or throwing errors.

PR #232, opened 2019-05-09
Add remodeling support for serial-request

glower:


service/sign/handlers_sign.go
```diff
@@ -133,6 +177,54 @@ func Serial(w http.ResponseWriter, r *http.Request) response.ErrorResponse {
...
+
+		// Check if find model maches requested model
+		if modelAssert.HeaderString("model") != substore.ModelName {
+			const msg = "Requested model is invalid"
+			svlog.Message("SIGN", "invalid-assertion", msg)
+			return response.ErrorResponse{Success: false, Code: response.ErrorInvalidAssertion.Code, Message: msg, StatusCode: http.StatusBadRequest}
+		}
+
+		// Don't know what to check for the original serial
+		if serialAssert.HeaderString("model") != originalModel {
```
pedronis: serial-request is self signed, I think/hope we have code to check that?  the model assertion we get is mostly for information ATM.

But for the old serial we need need to check it quite thoroughly, that it matches  all the original-* fields,   that the device key is the same between serial-request and old serial.  And also we need to actually check the signature of old serial,  it came from this vault under current assumptions so we should have the key that signed it around to check it.

service/sign/handlers_sign.go
```diff
@@ -133,6 +177,54 @@ func Serial(w http.ResponseWriter, r *http.Request) response.ErrorResponse {
...
+
+		// Validate the new model: it must be defind in the sub-store of the orignal model
+		substore, err := datastore.Environ.DB.GetSubstore(originalModelAssert.ID, originalSerial)
+		if err != nil {
+			svlog.Message("PIVOT", "invalid-substore", "Cannot find sub-store mapping for the model")
+			return response.ErrorInvalidSubstore
+		}
+
+		// Check if find model maches requested model
+		if modelAssert.HeaderString("model") != substore.ModelName {
```
pedronis: here we can also use serialReq, right,  we double checked the match before

glower:
Implemented additional checks for the original serial
Not sure if I did signature validation correct

service/sign/handlers_sign.go
```diff
@@ -133,15 +177,102 @@ func Serial(w http.ResponseWriter, r *http.Request) response.ErrorResponse {
...
+			return response.ErrorResponse{Success: false, Code: response.ErrorInvalidAssertion.Code, Message: msg, StatusCode: http.StatusBadRequest}
+		}
+		// Check that the device key is the same between serial-request and old serial
+		if serialAssert.HeaderString("device-key") != serialReq.HeaderString("device-key") {
+			const msg = "Device-key is invalid"
+			svlog.Message("SIGN", "invalid-assertion", msg)
+			return response.ErrorResponse{Success: false, Code: response.ErrorInvalidAssertion.Code, Message: msg, StatusCode: http.StatusBadRequest}
+		}
+
+		// Sign the sended old serial assertion and compare the signature with sended signature
```
pedronis: it should be possible to check this with something like:

err := asserts.SignatureCheck(serialAssert,  ...KeypairDB.PublicKey(KEY_ID))

service/sign/handlers_sign_test.go
```diff
@@ -74,6 +77,7 @@ func sendRequest(method, url string, data io.Reader, apiKey string, c *check.C)
 func (s *SignSuite) TestSerial(c *check.C) {
 	// Generate a test serial-request assertion
 	assert, err := generateSerialRequestAssertion("alder", "A123456L", "")
+	fmt.Printf("\n\n%s\n\n", assert)
```
pedronis: left over debug print?

service/sign/handlers_sign_test.go
```diff
@@ -328,3 +516,120 @@ func (s *SignSuite) TestSignHandlerErrorKeyStore(c *check.C) {
...
+func (s *SignSuite) TestRemodeling(c *check.C) {
+	// For the remodelling we need to send serial-request assertion + new model assertion + old serail assertion
+	// First, get the serial assertion for the original model
+	serialReq, err := generateSerialRequestAssertion("alder", "A123456L", "")
+	c.Assert(err, check.IsNil)
+	w := sendRequest("POST", "/v1/serial", bytes.NewReader(serialReq), "ValidAPIKey", c)
+	c.Assert(w.Code, check.Equals, 200)
+	c.Assert(w.Body, check.NotNil)
+	serialAssertions := w.Body.String()
+	fmt.Printf("\n\n%s\n\n", serialAssertions)
```
pedronis: left over debug print?

service/sign/handlers_sign.go
```diff
@@ -181,11 +253,91 @@ func Serial(w http.ResponseWriter, r *http.Request) response.ErrorResponse {
...
+	}
+
+	// Check that the device key is the same between serial-request and old serial
+	if serialAssert.HeaderString("device-key") != serialReq.HeaderString("device-key") {
+		const msg = "Device-key is invalid"
+		svlog.Message("SIGN", "invalid-assertion", msg)
+		return response.ErrorResponse{Success: false, Code: response.ErrorInvalidAssertion.Code, Message: msg, StatusCode: http.StatusBadRequest}
+	}
+
+	oldModelPublicKey, err := datastore.Environ.KeypairDB.PublicKey(substore.FromModel.KeyID)
```
pedronis: the  way to find the key id here is a bit indirect,  I think we should check that KeyID == serialAssert.HeaderString("sign-key-sha3-384") as well

service/sign/handlers_sign_test.go
```diff
@@ -328,3 +514,119 @@ func (s *SignSuite) TestSignHandlerErrorKeyStore(c *check.C) {
...
+		{false, "POST", "/v1/serial", wrongModel, 400, response.JSONHeader, "ValidAPIKey"},                 //  6
+		{false, "POST", "/v1/serial", assertionsBad, 400, response.JSONHeader, "ValidAPIKey"},              //  7
+		{false, "POST", "/v1/serial", assertionsWrongSerial, 400, response.JSONHeader, "ValidAPIKey"},      //  8
+		{false, "POST", "/v1/serial", assertionsBedSerialNumber, 400, response.JSONHeader, "ValidAPIKey"},  //  9
+		{false, "POST", "/v1/serial", assertionsWrongModel, 400, response.JSONHeader, "ValidAPIKey"},       // 10
+		{false, "POST", "/v1/serial", assertionsWrongDeviceKey, 400, response.JSONHeader, "ValidAPIKey"},   // 11
+		{false, "POST", "/v1/serial", assertionWrongSerialNumber, 400, response.JSONHeader, "ValidAPIKey"}, // 12
+		{false, "POST", "/v1/serial", assertionsWrongBrand, 400, response.JSONHeader, "ValidAPIKey"},       // 13
+		{false, "POST", "/v1/serial", assertionsWrongSignature, 400, response.JSONHeader, "ValidAPIKey"},   // 14
+		// {false, "POST", "/v1/serial", assertionsWrongSignKey, 400, response.JSONHeader, "ValidAPIKey"}, // 15
```
matiasb: Is this test missing ^?

glower: yes, added the test

PR #236, opened 2019-06-13
WIP: Add more information to the error message log

glower:
* Add detailed information about the error to the error response

### DIFF
--- a/service/sign/handlers_sign.go
+++ b/service/sign/handlers_sign.go
@@ -346,7 +346,8 @@ func findModel(brandID, modelName, serialNumer, apiKey string) (datastore.Model,
 	// Validate the model by checking that it exists on the database
 	model, err := datastore.Environ.DB.FindModel(brandID, modelName, apiKey)
 	if err != nil {
-		svlog.Message("SIGN", response.ErrorInvalidModel.Code, response.ErrorInvalidModel.Message)
+		msg := fmt.Sprintf("Cannot find model %s with the matching brand %s and apiKey %s", modelName, brandID, apiKey)
+		svlog.Message("SIGN", "invalid-model", msg)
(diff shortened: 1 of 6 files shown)


glower:
This is a possible fix for https://github.com/CanonicalLtd/serial-vault/issues/235
If this PR is OK I will migrate rest of the responses to have more information

service/sign/handlers_sign.go
```diff
@@ -346,7 +346,8 @@ func findModel(brandID, modelName, serialNumer, apiKey string) (datastore.Model,
 	// Validate the model by checking that it exists on the database
 	model, err := datastore.Environ.DB.FindModel(brandID, modelName, apiKey)
 	if err != nil {
```
pedronis: because of the try FindModel and then otherwise on failure eagerly try GetSubstoreModel it seems - as it was already - this will always produce slightly odd error messages/responses,  this isn't new but if the goal is making for more understandable errors for the brand user this might need a larger rethink

glower: OK, I will drop this PR and only put more info inside of the internal log. We can have a separate conversation about error responses.

pedronis: I would ask @cprov whether the change here is already enough of an improvement, he was the one having trouble debugging, before dropping them.  For more clarity the issue here is how apiKey is checked implicitly in one case but not the other, I don't know the detail of the data model and other usages enough but in principle it would be better if the apiKey check was explicit/distinguishable unless we think leaking whether a model exist or not is deemed a problem.

glower: Yes, I agree on that, but in this case, I just wanted to improve logging, not touching any business logic at all.

PR #238, opened 2019-06-19
Fixed generated key registration in the store

glower:
This is a fix for https://github.com/CanonicalLtd/serial-vault/issues/234

As I can tell from the previous commits in this module, `keypairstatus` table is only used temporarily during key generation form the UI. After the key is generated the entry from `keypairstatus` is removed, this is why 
`getKeypairByNameSQL` query was not working.

As I can tell `key_name` was moved from `keypairstatus` to `keypair` so we don't need the JOIN on `keypairstatus` table anymore.

Tested manually on my local machine and the error is not appearing anymore. Not sure if key registration in the store is working, but particular this bug is fixed.


glower:
@slimjim777 please have a short look for PR

PR #242, opened 2019-10-22
Change regex for username validation

glower:


glower:
User name validation will be changed to any string without blanks

PR #250, opened 2019-12-13
Return correct error message UpsertModelAssert function

glower:
The problem was: if you create a new model and hit "Configure the model assertion headers" to create or edit an assertion for this model, you will always get an error message: "Error creating the assertion: error upserting the model assertion for model 26: <nil>" even if all was good


glower:
No, there are no tests for this part at all

PR #255, opened 2020-01-29
Fixing signinglog search, frontend part

glower:
At the moment we don't have an automated build for the frontend code, so I needed to build `static/js/bundle.js` by myself. 

The tests are broken on master and we don't have any automation to run the frontend tests for each build (probably this is why the tests are broken). So I did not touch the tests at all, because then I need to fix all the tests.

We definitely need to make some housekeeping for the front-end part of this project but not with the current PR.


glower:
Feel free to comment/ask for the change on everything, it's kinda my first react.js code (I did some node on the backend before).
This is frontend change only, the change for the backend will be created next, so it is easy to review the changes. We don't have automation deploy, so this change will not break some live service

glower:
Hmm... I have no idea why Travis is unhappy, I will investigate this. This has nothing to do with the current change, it only checks the backend code

doismellburning:
> At the moment we don't have an automated build for the frontend code, so I needed to build static/js/bundle.js by myself.

Thoughts on addressing that first?

It feels worth prioritising getting something known-good/working automatically building, to be able to validate all future changes - rather than making those changes somewhat "blind", as it were

glower:
I will have a look into it first. We need to fix the tests, run it by Travis, run the build and then let Travis commit the build.

glower:
- Frontend tests are fixed and running in the CI
- static/js/bundle.js will be auto-generated by the CI (this will be done later)

webapp-admin/src/components/SigningLog.js
```diff
@@ -30,12 +30,20 @@ class SigningLog extends Component {
...
+        filterString: '',
         message: null,
         expanded: {models: true},
         query: '',
         startRow: 0,
         endRow: PAGINATION_SIZE,
+        totalLogs: 0,
+        authorityID: null,
     };
+    this.state.authorityID = this.props.selectedAccount.AuthorityID
```
Lukewh: you could change line 42 to:
`authorityID: this.props.selectedAccount.AuthorityID`
and remove this line

webapp-admin/src/components/SigningLog.js
```diff
@@ -45,33 +53,48 @@ class SigningLog extends Component {
...
+      items[index].selected = !items[index].selected;
+      var filter = items.filter(i => i.selected === true).map(i => i.name).join(',')
+      this.getSigningLogs(this.state.authorityID, 0, filter, this.state.query)
+      this.setState({filterModels: items, filterString: filter});
+    }
   }
 
-  handleRecordsForPage = (startRow, endRow) => {
-    this.setState({startRow: startRow, endRow: endRow});
+  getSigningLogs = (authorityID, offset, filter, serialnumber) => {
```
Lukewh: If `authorityID` is always `this.state.authorityID` you can change these args to:
`getSigningLogs = (offset, filter, serialnumber)`
Then get the state within the function:
`var authorityID = this.state.authorityID;`
That way you don't need to pass the state from every call to this.

webapp-admin/src/components/SigningLog.js
```diff
@@ -45,33 +53,48 @@ class SigningLog extends Component {
...
+      var data = JSON.parse(response.body);
+      var message = null;
+      if (!data.success) {
+        message = data.message;
+      }
+      this.setState({logs: data.logs, message: message, totalLogs: data.total});
+    });
+  }
+
+  getSigningLogFilters = (authorityID) => {
```
Lukewh: Same here as with `getSigningLogs`, you don't need to pass `authorityID` if it's always from the state

webapp-admin/src/components/SigningLog.js
glower: fixed

webapp-admin/src/components/SigningLog.js
glower: fixed

webapp-admin/src/components/SigningLog.js
glower: fixed

glower:
Thank you @Lukewh 
I did the QA with my backend changes so I am confident it works

PR #256, opened 2020-01-29
Fixing signing log search, backend part

glower:
This PR changes the way we search in the signing logs. Currently, frontend will get the first 10k entries and do the search and filtering in the frontend, based on this 10k data. So the main bug is there is no way to find the data if someone has more than 10k entries.

The main change is to make the pagination, search and filtering in the database. The change in the frontend code is done in the separate PR.

I tried to make changes for the code as small as possible. It is just passing through the params from the frontend to the backend and create a SQL query based on the params. For this, I introduced SQL query builder to make it easier.


datastore/signinglogmanager.go
```diff
@@ -71,18 +76,6 @@ const listSigningLogForUserSQL = `
 	)
 	ORDER BY id DESC LIMIT 10000`
 
-const listSigningLogForAccountSQL = "SELECT * FROM signinglog WHERE id < $1 AND make=$2 ORDER BY id DESC LIMIT 10000"
```
glower: these two queries are migrated to the query builder

service/signinglog/actions.go
```diff
@@ -37,6 +37,7 @@ type ListResponse struct {
 	ErrorSubcode string                 `json:"error_subcode"`
 	ErrorMessage string                 `json:"message"`
 	SigningLog   []datastore.SigningLog `json:"logs"`
```
pedronis: afaiu these will containt total_count as well? that's a bit strange, that could be made at least not JSON serialized in SigningLog itself

.travis.yml
```diff
@@ -19,6 +19,7 @@ before_install:
   - go get github.com/client9/misspell/cmd/misspell
   - go get -u github.com/gordonklaus/ineffassign
   - go get -u github.com/alexkohler/nakedret
+  - go get github.com/lann/builder
```
pedronis: how is this used?  it's seem unrelated to the main change here

glower: this is somehow the dependency from `github.com/Masterminds/squirrel` but govendor didn't manage to add this to the main vendor.json file, so I added it here.

service/signinglog/actions.go
glower: yes, this is correct, I just wanted to have the counter on the top level of the data

service/signinglog/actions.go
```diff
@@ -37,7 +37,6 @@ type ListResponse struct {
...
-	Total        int                    `json:"total"`
```
pedronis: I don't understand this change,  my suggestion talked about SigningLog, what I was thinking was more:

```
@@ -110,6 +103,7 @@ type SigningLog struct {
	Created      time.Time `json:"created"`
	Revision     int       `json:"revision"`
	Synced       int       `json:"synced"`
	Total        int       `json:"total_count"`
}
```

remove the `json:"total_count"` tag there

PR #258, opened 2020-01-29
push golang version to 1.13 in travis

glower:


.travis.yml
```diff
@@ -15,7 +15,7 @@ before_install:
     - go get gopkg.in/retry.v1
     - go get gopkg.in/tomb.v2
     - go get gopkg.in/check.v1
-    - go get github.com/golang/lint/golint
+    - go get -u golang.org/x/lint/golint
```
glower: this is the new repository of the linter

glower:
The golang linter we are using is not working with the older golang (1.9) version. So to be able to run travis builds we need to update to the latest stable version 1.13

PR #261, opened 2020-02-07
Fixed wrong param to get the signing log

glower:


webapp-admin/src/components/SigningLog.js
```diff
@@ -69,7 +69,7 @@ class SigningLog extends Component {
       if (!data.success) {
         message = data.message;
       }
-      this.setState({logs: data.logs, message: message, totalLogs: data.total});
+      this.setState({logs: data.logs, message: message, totalLogs: data.total_count});
```
glower: we changed this name in the backend, so I need to update it here

webapp-admin/src/components/SigningLog.js
```diff
@@ -57,7 +57,7 @@ class SigningLog extends Component {
       items = this.state.filterModels;
       items[index].selected = !items[index].selected;
       var filter = items.filter(i => i.selected === true).map(i => i.name).join(',')
-      this.getSigningLogs(this.state.authorityID, 0, filter, this.state.query)
```
glower: as I updated the method to work without the authorityID I forgot to fix all the calls of the method :(

PR #267, opened 2020-02-25
Added github action for the automated frontend build

glower:


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,29 @@
+name: Automated javascript build CI
+on:
+  push:
+    paths: 'webapp-admin/**'
```
glower: This will trigger only on changes in `webapp-admin` and all the sub-directories of this

webapp-admin/package.json
```diff
@@ -28,5 +28,17 @@
     "node-sass": "^4.12.0",
     "react-dom": "^16.12.0",
     "vanilla-framework": "^1.7.1"
+  },
+  "browserslist": {
```
glower: this was automatically added by running `npm run build` so I assume it's OK here

webapp-admin/public/index.html
```diff
@@ -7,15 +7,11 @@
 	<title>{{.Title}}</title>
 
 	<link rel="stylesheet" href="/static/font-awesome/css/font-awesome.min.css" />
-	<link rel="stylesheet" href="/static/css/application.css" />
```
glower: the new built js/css files look like `main.652bcdd4.chunk.js` and are automatically added to this html file. So we don't need these hardcoded static names here.

glower:
> Looks good. It's using create-react-app right? That's why the relevant places in the HTML get swapped out? At least that's what I'm assuming.

Yes, this is correct, it's create-react-app 
Thanks for having a look

PR #272, opened 2020-03-12
Fixed permissions so an admin users can't create an account anymore

glower:
Only users with SuperAdmin rights are allowed to create new accounts. This means that the `Account` tab in the admin UI is not visible to all other users anymore (same logic as for the `User` tab).

### DIFF
--- a/service/account/handlers_accounts_test.go
+++ b/service/account/handlers_accounts_test.go
@@ -153,25 +153,31 @@ func (s *AccountSuite) TestCreateGetUpdateAccountHandlers(c *check.C) {
 
 	account := datastore.Account{ID: 2, AuthorityID: "vendor", ResellerAPI: false}
 	acc, _ := json.Marshal(account)
+	accBrokenJSON := []byte(`{"user"`)
 
 	tests := []AccountTest{
-		{"POST", "/v1/accounts", nil, 400, "application/json; charset=UTF-8", 0, false, false, false, false, 0},
-		{"POST", "/v1/accounts", acc, 200, "application/json; charset=UTF-8", 0, false, true, false, false, 0},
-		{"POST", "/v1/accounts", acc, 200, "application/json; charset=UTF-8", datastore.Admin, true, true, false, false, 1},
-		{"POST", "/v1/accounts", acc, 400, "application/json; charset=UTF-8", datastore.Admin, true, false, true, false, 1},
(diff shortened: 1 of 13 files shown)


service/account/handlers_accounts_test.go
```diff
@@ -153,25 +153,31 @@ func (s *AccountSuite) TestCreateGetUpdateAccountHandlers(c *check.C) {
 
 	account := datastore.Account{ID: 2, AuthorityID: "vendor", ResellerAPI: false}
 	acc, _ := json.Marshal(account)
+	accBrokenJSON := []byte(`{"user"`)
 
 	tests := []AccountTest{
```
doismellburning: This is not especially easy or pleasant to review and I'm surprised by the sheer number of changes here, including ones that don't seem obviously relevant

Thoughts on splitting it up into single-change commits / adding some more clarity in some other way?

doismellburning:
The Go changes certainly look good (though I struggle with the test changes)!

The JS seems reasonable at a glance but I don't really JS...

service/account/handlers_accounts_test.go
glower: I added some missing test cases, but you are right, I will remove all not necessary test cases from this PR (but yes, the way the tests are structured makes it hard to fix/review)

glower:
The changes in the frontend code are removing the navigation tab `Accounts` from the UI if the user does not have `SuperUser` status

glower:
> The Go changes certainly look good (though I struggle with the test changes)!
Removed all not necessary changes in the tests. For the better overview of the changes, I need to reformat all the tests to not to use the short form for the map. But this change is just too large 

> The JS seems reasonable at a glance but I don't really JS...
I will ask for the extra FE code review

webapp-admin/src/components/AccountEdit.js
```diff
@@ -37,7 +37,7 @@ class AccountEdit extends Component {
     getAccount(accountId) {
         Accounts.get(accountId).then((response) => {
             var data = JSON.parse(response.body);
-
+
```
Lukewh: Cheeky spaces

service/account/handlers_accounts_test.go
doismellburning: Thanks! Definitely agree that these tests _generally_ aren't easy to review, but it's much clearer now you've dropped the extraneous ones cheers!

PR #275, opened 2020-03-18
Clean up the previews frontend build before create a new one

glower:
`npm run build`  creates unique css/js files with hashes as part of the filename. Because of this, we need the remove all css/js files from the previews build

### DIFF
--- a/webapp-admin/package.json
+++ b/webapp-admin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webapp-admin",
-  "version": "0.4.0",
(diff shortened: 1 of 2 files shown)


webapp-admin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "webapp-admin",
-  "version": "0.4.0",
+  "version": "0.4.1",
```
glower: bump the FE version to trigger the new build

PR #276, opened 2020-03-18
Migrating the project to Go Modules

glower:
* Moving from `govendor` which is deprecated now to `go mod` for the dependency management. All the external dependencies have the same version now, except `snapd`
* Snapd version is at 2.44 now, all older versions were not compatible with Go mod
* Updated documentation about how to use Go mod
* Updated travis configuration


glower:
WIP
Following the steps from the official post: https://blog.golang.org/migrating-to-go-modules
Running `go mod init github.com/CanonicalLtd/serial-vault` copied all the pinned dependencies from `vendor/vendor.json` to `go.mod`

The only issue was with the older version of `snapd` this problem was solved in one of the letest versions, so I migrated to snapd `2.44` with
```
go get github.com/snapcore/snapd@2.44
```

account/account.go
```diff
@@ -32,8 +32,8 @@ import (
 // FetchAssertionFromStore retrieves an assertion from the store
 var FetchAssertionFromStore = func(modelType *asserts.AssertionType, headers []string) (asserts.Assertion, error) {
 	var user *auth.UserState
-	var authContext auth.AuthContext
-	sto := store.New(nil, authContext)
+	var storeCtx store.DeviceAndAuthContext
```
glower: The change to move `AuthContext` to a different modue: https://github.com/snapcore/snapd/commit/30311dbee2e11ec01f34c1c822d87e756d966e6d

This is the only API change from `snapd` I needed to add because of the new version of snapd

go.mod
```diff
@@ -0,0 +1,36 @@
+module github.com/CanonicalLtd/serial-vault
+
+go 1.13
+
+require (
+	github.com/Masterminds/squirrel v1.2.0
+	github.com/dgrijalva/jwt-go v3.2.0+incompatible
+	github.com/godbus/dbus v4.1.0+incompatible // indirect
+	github.com/gorilla/csrf v1.0.3-0.20161122164500-69581736821c
```
doismellburning: In `vendor.json` there was `"versionExact": "v1.5"` - how come the change?

I suspect there are other differences like this but I'm not doing a full audit. If they're necessary/reasonable/ok then sure, but provide context in e.g. the commit message or something please

glower: `v1.5` is not a valid SemVer as specified in the Go mod spec (it need to be `v1.5.0`) so it will be somehow translated into the date and hash value of the latest commit from this tag.

It's the same version, I double-checked every direct dependency.
https://github.com/gorilla/csrf/releases/tag/v1.5 has the hash ` 6958173` and the date is `Nov 22, 2016` (this is the date and hash from the latest commit in this version).

glower: From the vendor.json: the `revision` and `revisionTime` is the same as in go.mod. So the same git commit is pinned here and there.
```
		{
			"checksumSHA1": "2ieXR7YGRBDmQWVeUpjk5ay8iCE=",
			"path": "github.com/gorilla/csrf",
			"revision": "69581736821c33d85bbf378f42f6ad864dbd85de",
			"revisionTime": "2016-11-22T16:45:00Z",
			"version": "v1",
			"versionExact": "v1.5"
		},
```

doismellburning: Makes sense, cheers - want to actually put this in the change so it's more visible for anyone else?

glower: OK, I will put this change into ChangeLog file and then in release notes in GitHub

glower: Added a note in the ChangeLog file

PR #281, opened 2020-03-23
Added new middleware for the prometheus metrics

glower:


go.mod
```diff
@@ -19,13 +19,14 @@ require (
...
 	github.com/pkg/errors v0.8.1-0.20180311214515-816c9085562c
+	github.com/prometheus/client_golang v1.1.0
 	github.com/snapcore/bolt v1.3.1 // indirect
 	github.com/snapcore/go-gettext v0.0.0-20191107141714-82bbea49e785 // indirect
 	github.com/snapcore/snapd v0.0.0-20200317200833-16631e228c07
 	github.com/yohcop/openid-go v0.0.0-20170901155220-cfc72ed89575
-	golang.org/x/crypto v0.0.0-20180409150046-beb2a9779c3b
-	golang.org/x/net v0.0.0-20180406214816-61147c48b25b
-	golang.org/x/sys v0.0.0-20180406135729-3b87a42e500a
+	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
```
glower: These libs are updated automatically because of introducing of `prometheus/client_golang`

go.sum
```diff
@@ -1,10 +1,26 @@
 github.com/Masterminds/squirrel v1.2.0 h1:K1NhbTO21BWG47IVR0OnIZuE0LZcXAYqywrC3Ko53KI=
 github.com/Masterminds/squirrel v1.2.0/go.mod h1:yaPeOnPG5ZRwL9oKdTsO/prlkPbXWZlRVMQ/gGlzIuA=
```
glower: ignore this file, it is automatically generated by `go mod` tool adding `prometheus/client_golang`

service/metric/handler.go
```diff
@@ -0,0 +1,27 @@
...
+func NewServer() *Server {
+	return &Server{
+		metrics: promhttp.Handler(),
+	}
+}
+
+// ServeHTTP returns a metrics server.
+// Use it to set up the prometheus endpoint in your web service
+// router := mux.NewRouter()
+// router.Handle("/_status/metrics", metric.NewServer()).Methods("GET")
```
glower: I created an extra package only for the `http.Handler` so we can easily integrate the monitoring endpoint not depending on which HTTP router or framework is used.

service/metric/metrics.go
```diff
@@ -0,0 +1,50 @@
+package metric
```
glower: we use the same basic metrics in `snapstorage` and other store projects, so for consistency reasons, I left naming and labels the same.

service/metric/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+	"net/http"
+	"strconv"
+	"time"
+)
+
+// CollectAPIStats middleware collects different statistics from the HTTP request
+// use this middleware like this:
+// router := mux.NewRouter()
+// router.Handle("/v1/models", metric.CollectAPIStats("modelList", http.HandlerFunc(model.List))).Methods("GET")
+func CollectAPIStats(view string, inner http.Handler) http.Handler {
```
glower: This is the basic HTTP middleware to collect some stats from the HTTM request/response. It can be added to every HTTP router

service/metric/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		// record prometheus metric for latency in ms
+		latency := float64(time.Since(start).Nanoseconds()) / 1000000
+
+		HTTPIncomingLatencyHistogramVec.WithLabelValues(r.Method, ww.Status(), view).Observe(latency)
+		HTTPIncomingRequestCounterVec.WithLabelValues(r.Method, ww.Status(), view).Inc()
+	})
+}
+
+// recordResponse is a proxy around an http.ResponseWriter
```
glower: This is the trick to get a response status from the already written response. `bytesWritten` are not used now, but we can add an additional metric for this information

service/metric/middleware_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var expectedPrometheusData = map[string]string{
```
glower: the same test I did for the snapstorage

glower:
This is only the implementation of the middleware and some basic metrics we have for all other services in the store. I will do the integration in the separate PR.

This is a basic bootstrap to add additional metrics in an easy way for everything else we wish to measure.

service/metric/metrics.go
```diff
@@ -0,0 +1,50 @@
...
+		Buckets: []float64{4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192},
+	},
+	[]string{"method", "status", "view"},
+)
+
+// HTTPIncomingErrorsCounterVec is prometheus metric for HTTP errors
+var HTTPIncomingErrorsCounterVec = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "http_in_errors",
+		Help: "fetric for HTTP errors",
```
doismellburning: metric

service/metric/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+		// count all server side errors
+		if ww.status >= 500 {
+			HTTPIncomingErrorsCounterVec.WithLabelValues(r.Method, ww.Status(), view).Inc()
+		}
+		// count 504 Gateway Timeout
+		if ww.status == 504 {
+			HTTPIncomingTimeoutsCounterVec.WithLabelValues(r.Method, view).Inc()
+		}
+		// record prometheus metric for latency in ms
+		latency := float64(time.Since(start).Nanoseconds()) / 1000000
```
doismellburning: https://golang.org/pkg/time/#Duration.Milliseconds ?

doismellburning:
Not APIs / functionality I have significant familiarity with, but it all looks reasonable, thanks

service/metric/middleware.go
glower: right, `Milliseconds` was added in `1.13`, will fix it here

PR #282, opened 2020-03-31
Integrated prometheus middleware to collect metrics for HTTP calls

glower:
Added basic Prometheus metrics for HTTP stats to every route

### DIFF
--- a/service/metric/handler.go
+++ b/service/metric/handler.go
@@ -2,6 +2,7 @@ package metric
 
 import (
 	"net/http"
+	"sync"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
@@ -11,8 +12,13 @@ type Server struct {
 	metrics http.Handler
 }
 
+var doOnce sync.Once
+
 // NewServer returns a new metrics server.
 func NewServer() *Server {
+	doOnce.Do(func() {
+		InitMetrics()
+	})
 	return &Server{
 		metrics: promhttp.Handler(),
 	}
--- a/service/router.go
+++ b/service/router.go
@@ -205,29 +205,29 @@ func AdminRouter() *mux.Router {
 		Methods("POST")
 
 	// API routes: users management
-	router.Handle("/v1/users", metric.CollectAPIStats("XXX",
+	router.Handle("/v1/users", metric.CollectAPIStats("userList",
 		MiddlewareWithCSRF(http.HandlerFunc(user.List)))).
 		Methods("GET")
-	router.Handle("/v1/users", metric.CollectAPIStats("XXX",
+	router.Handle("/v1/users", metric.CollectAPIStats("userCreate",
 		MiddlewareWithCSRF(http.HandlerFunc(user.Create)))).
 		Methods("POST")
-	router.Handle("/v1/users/{id:[0-9]+}", metric.CollectAPIStats("XXX",
+	router.Handle("/v1/users/{id:[0-9]+}", metric.CollectAPIStats("userGet",
 		MiddlewareWithCSRF(http.HandlerFunc(user.Get)))).
 		Methods("GET")
-	router.Handle("/v1/users/{id:[0-9]+}", metric.CollectAPIStats("XXX",
+	router.Handle("/v1/users/{id:[0-9]+}", metric.CollectAPIStats("userUpdate",
 		MiddlewareWithCSRF(http.HandlerFunc(user.Update)))).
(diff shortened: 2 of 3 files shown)


service/metric/handler.go
```diff
@@ -11,8 +12,13 @@ type Server struct {
 	metrics http.Handler
 }
 
+var doOnce sync.Once
+
 // NewServer returns a new metrics server.
 func NewServer() *Server {
+	doOnce.Do(func() {
```
glower, line 19: This is needed to run `InitMetrics()` once, because otherwise, Prometheus will complain about metrics registration. It only happens if we run tests. In normal case, this code will be executed once anyway.
.
In Golang you can reach the same using `init()` function. The code int this function will be executed only once on the import. But this is somehow magic and I don't like to use `init()`.

service/router.go
```diff
@@ -47,23 +48,41 @@ func SigningRouter() *mux.Router {
...
 	router.Handle("/v1/version", Middleware(http.HandlerFunc(core.Version))).Methods("GET")
 	router.Handle("/v1/health", Middleware(http.HandlerFunc(core.Health))).Methods("GET")
-	router.Handle("/v1/serial", Middleware(ErrorHandler(sign.Serial))).Methods("POST")
-	router.Handle("/v1/request-id", Middleware(ErrorHandler(sign.RequestID))).Methods("POST")
-	router.Handle("/v1/model", Middleware(ErrorHandler(assertion.ModelAssertion))).Methods("POST")
-	router.Handle("/v1/pivot", Middleware(ErrorHandler(pivot.Model))).Methods("POST")
-	router.Handle("/v1/pivotmodel", Middleware(ErrorHandler(pivot.ModelAssertion))).Methods("POST")
-	router.Handle("/v1/pivotserial", Middleware(ErrorHandler(pivot.SerialAssertion))).Methods("POST")
-	router.Handle("/v1/pivotuser", Middleware(ErrorHandler(pivot.SystemUserAssertion))).Methods("POST")
+
```
glower, line 53: the only change in this file was to add `metric.CollectAPIStats("callerName" ...` to every route we have here. And some reformating to not have the line too long

glower:
I will create an additional PR and document all the metrics we collect so far

service/router.go
```diff
@@ -47,23 +48,41 @@ func SigningRouter() *mux.Router {
...
+		Methods("POST")
 
 	// Test log upload routes (only in the factory)
 	if datastore.InFactory() {
 		router.Handle("/testlog", Middleware(http.HandlerFunc(testlog.Index))).Methods("GET")
 		router.Handle("/testlog", Middleware(http.HandlerFunc(testlog.Submit))).Methods("POST")
 	}
 
+	// prometheus metrics endpint
+	router.Handle("/_status/metrics", metric.NewServer()).Methods("GET")
```
glower, line 84: Serial Vault runs as two different services, once for API and once for the Admin/UI. This is why we need to have this endpoint twice

service/router.go
```diff
@@ -76,62 +95,140 @@ func AdminRouter() *mux.Router {
...
+		MiddlewareWithCSRF(http.HandlerFunc(user.Delete)))).
+		Methods("DELETE")
+	router.Handle("/v1/users/{id:[0-9]+}/otheraccounts", metric.CollectAPIStats("XXX",
+		MiddlewareWithCSRF(http.HandlerFunc(user.GetOtherAccounts)))).
+		Methods("GET")
 
 	// OpenID routes: using Ubuntu SSO
-	router.Handle("/login", MiddlewareWithCSRF(http.HandlerFunc(usso.LoginHandler)))
-	router.Handle("/logout", MiddlewareWithCSRF(http.HandlerFunc(usso.LogoutHandler)))
+	router.Handle("/login", metric.CollectAPIStats("XXX",
```
doismellburning, line 228: Either you've missed something here with the `XXX` or it could be clearer that this is intentional please!

glower: Omg, it was the placeholder I missed, fixing

glower:
@bloodearnest normally this is the way how it's done with the HTTP middleware in Golang (at least with the current router we use). So you normally always add the middleware to the route explicit.HTTP routes declaration in Golang usually looks like this: 
https://github.com/drone/drone/blob/master/handler/api/api.go
https://github.com/minio/minio/blob/6f6a2214fc0294569a8e2c992d23db1c4f3c5851/cmd/api-router.go

PR #283, opened 2020-04-07
Added endpoints for liveness check and the database check

glower:


glower:
We do have similar endpoints in Serial Vault, but they are not consistent with our existing check endpoints. To not change existing implementation I added to new endpoints.

In the next PR I will add the documentation for all the endpoints we have in `/_status/`

PR #287, opened 2020-04-28
WIP: remove single quotes and leading and trailing white spaces

glower:
remove single quotes and leading and trailing white spaces from the header

### DIFF
--- a/service/sign/handlers_sign.go
+++ b/service/sign/handlers_sign.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/CanonicalLtd/serial-vault/service/log"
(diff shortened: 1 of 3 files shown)


service/sign/handlers_sign.go
```diff
@@ -254,10 +255,22 @@ func Serial(w http.ResponseWriter, r *http.Request) response.ErrorResponse {
...
+	header = strings.Replace(header, "'", "", -1)
+	header = strings.TrimSpace(header)
+	return header
+}
+
 func checkRemodelingRequest(serialReq *asserts.SerialRequest, modelAssert, serialAssert asserts.Assertion, apiKey string) response.ErrorResponse {
-	originalBrandID := serialReq.HeaderString("original-brand-id")
-	originalModel := serialReq.HeaderString("original-model")
-	originalSerial := serialReq.HeaderString("original-serial")
+	originalBrandID := CleanHeader(serialReq.HeaderString("original-brand-id"))
```
slimjim777, line 266: CleanHeader isn't really needed for the original brandID and original model as they are coming cleanly from the assertions

service/sign/handlers_sign.go
```diff
@@ -254,10 +255,22 @@ func Serial(w http.ResponseWriter, r *http.Request) response.ErrorResponse {
 	return response.ErrorResponse{Success: true}
 }
 
+// CleanHeader removes single quotes and leading and trailing white spaces from the header
+func CleanHeader(header string) string {
+	header = strings.Replace(header, "'", "", -1)
```
slimjim777, line 260: This looks fine - there are TrimPrefix and TrimSuffix that will remove the leading and trailing characters. But I think this is fine.

service/sign/handlers_sign_test.go
```diff
@@ -660,3 +661,26 @@ func (s *SignSuite) TestRemodeling(c *check.C) {
...
 }
+
+func (s *SignSuite) TestCleanHeader(c *check.C) {
+	tests := []struct {
+		name   string
+		header string
+		want   string
+	}{
+		{
+			name:   "case-1",
```
slimjim777, line 672: Also add a no-op case where there are no apostrophes.

service/sign/handlers_sign.go
glower: Make sense, I removed all the not necessary changes

slimjim777: Looks good

glower:
Serial number has single quotes in the serial-request assertion
The remodel request creates a serial-request assertion with some extra
headers. These are checked in the serial-vault. We've seen that
snapd is generating a serial-request assertion with the serial number
in single quotes e.g. original-serial: 'modelA'. That's probably
coming from the gadget and has been a long-standing issue.

The serial-vault then does a look-up for a "sub-store model" record
and fails because of the quotes. We've bypassed this by putting quotes
on the sub-store model record, but that should be a simple fix to
remove the quotes before the lookup.

PR #294, opened 2020-08-11
Replaced react-datepicker with native HTML datepicker

glower:
- Replaced react-datepicker with native HTML datepicker on Create System-User page
- Replaced moment.js with native js date object
- Added missing tests for "Create System-User" page
- Added golang test for the date-time format

### DIFF
--- a/webapp-admin/package.json
+++ b/webapp-admin/package.json
@@ -8,7 +8,6 @@
     "moment": "^2.24.0",
     "npm": "^6.13.6",
     "react": "^16.12.0",
-    "react-datepicker": "^2.11.0",
     "react-intl": "^3.11.0",
     "react-scripts": "^3.3.1",
     "then-request": "^6.0.2"
--- a/webapp-admin/src/components/SystemUserForm.js
+++ b/webapp-admin/src/components/SystemUserForm.js
@@ -15,27 +15,31 @@
  *
  */
 import React, {Component} from 'react'
-import moment from 'moment'
 import AlertBox from './AlertBox'
 import Assertion from '../models/assertions'
 import {T, isUserStandard} from './Utils'
 
-import DatePicker from 'react-datepicker'
-import 'react-datepicker/dist/react-datepicker-cssmodules.css'
+function newUTCDate() {
+    return new Date(new Date().toUTCString())
+}
 
 class SystemUserForm extends Component {
 
     constructor(props) {
         super(props);
 
+    	const startDate = newUTCDate();
+    	const endDate = new Date(newUTCDate().setFullYear(
+            startDate.getFullYear() + 1
+        ));
         this.state = {
             email: '',
             username: '',
             password: '',
             name: '',
             model: 0,
-            since: moment.utc(),
-            until: moment.utc().add(1, 'year'),
+            since: startDate,
+            until: endDate,
             message: '',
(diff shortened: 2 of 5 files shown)


glower:
After bumping the version of react and react-datepicker we broke /systemuser page. This happened because react-datepicker was not using moment.js anymore (https://www.npmjs.com/package/react-datepicker#momentjs)

With this change we will use native js Date object to fix the broken datepicker.
The change will also add missing test for this page

webapp-admin/src/components/SystemUserForm.js
```diff
@@ -21,19 +21,28 @@ import Assertion from '../models/assertions'
...
-import 'react-datepicker/dist/react-datepicker-cssmodules.css'
+import 'react-datepicker/dist/react-datepicker.css'
 
 class SystemUserForm extends Component {
 
     constructor(props) {
         super(props);
 
+    	// same as moment.utc()
+    	let startDate = new Date(new Date().toUTCString());
```
anthonydillon, line 32: These can be `const` as they are not reassigned.

webapp-admin/src/components/SystemUserForm.js
```diff
@@ -21,19 +21,28 @@ import Assertion from '../models/assertions'
...
+    	// same as moment.utc().add(1, "year")
+    	let endDate = new Date(new Date().toUTCString()).setFullYear(
```
anthonydillon, line 34: Same here

webapp-admin/src/components/SystemUserForm.js
```diff
@@ -215,7 +224,7 @@ class SystemUserForm extends Component {
                         <label htmlFor="since">Since (UTC):
                             <div className="row">
                                 <div className="col-3">
-                                    <DatePicker selected={this.state.since} onChange={this.handleChangeSinceDate} />
+                                    <DatePicker selected={this.state.startDate} onChange={this.handleChangeSinceDate} />
```
anthonydillon, line 227: Since handleChangeSinceDate function is updating state.since the selected date will not be updated unless this works by being the last thing the user selected.

glower:
* <DatePicker> + <input type="time">  is replaced with <input type="datetime-local"> element where you can set both: date and time with one element.
* onChange methods are now more simple

webapp-admin/src/components/SystemUserForm.js
glower: fixed

glower:
> Sorry I'm not sure what is the acturally problem here but was asked if someone could take a look. Would it be easier to use the native HTML datepicker if your not using moment:
> 
> ```
> <input type="date">
> ```

Thanks for the hint, I just replaced it with the native datepicker, it make more sense now and the code is simple now. 
I also added a backend test to test that the new generated format works there (in the end it's the same date-time string)

webapp-admin/package.json
```diff
@@ -8,7 +8,6 @@
     "moment": "^2.24.0",
     "npm": "^6.13.6",
     "react": "^16.12.0",
-    "react-datepicker": "^2.11.0",
```
anthonydillon, line 11: :+1:

anthonydillon: Always nice to remove deps

PR #296, opened 2020-08-12
Fix the system-user assertion form

slimjim777:
Fixes the loading of the system-user assertion form.
Adds the ssh-keys option to the form.


slimjim777:
This has been fixed on PR #294

PR #299, opened 2020-08-13
Replaced <a href=> with <button> elements

glower:
Having empty href attribute produces warnings during CI build process and stops the build process: "Treating warnings as errors because process.env.CI = true."

For the error: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid

### DIFF
--- a/webapp-admin/src/__tests__/dialogbox-test.js
+++ b/webapp-admin/src/__tests__/dialogbox-test.js
@@ -43,7 +43,7 @@ describe('dialog box', function() {
   // Check all the expected elements are rendered
   var divs = ReactTestUtils.scryRenderedDOMComponentsWithTag(page, 'div');
   expect(divs.length).toBe(2);
-  var anchors = ReactTestUtils.scryRenderedDOMComponentsWithTag(page, 'a');
+  var anchors = ReactTestUtils.scryRenderedDOMComponentsWithTag(page, 'button');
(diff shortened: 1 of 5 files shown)


glower:
The CI build error:
```
Treating warnings as errors because process.env.CI = true.
Most CI servers set it automatically.

Failed to compile.

./src/components/SubstoreList.js
  Line 135:21:  The href attribute is required for an anchor to be keyboard accessible. Provide a valid, navigable address as the href value. If you cannot provide an href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid


npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! webapp-admin@0.4.1 build: `npm run build-css && react-scripts build`
npm ERR! Exit status 1
```

glower:
Unfortunate it's not working. According to https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md `<a href="#" />` is not a valid ancor element. 
But in the UI it's a button, not a link, so it's kinda OK

PR #305, opened 2020-09-07
Added ols-vms config and additional Makefile target to build tarball

glower:
To be able to build and deploy serial-vault in the same way we deploy other stores services we need to have `ols-vms.conf` file in the root of the project.

Also added a Makefile target to build a tarball with binaries and static web files. So we can use only the Makefile during the build process and can get rid of some internal scripts.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -5,7 +5,7 @@ on:
     branches: master
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js 12.x
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,7 @@ TAR_BASE_NAME = ${SERVICE_NAME}-v${VERSION}.tar
 TGZ_NAME = ${TMP}/${SERVICE_NAME}-payload-${VERSION}.tgz
 SWIFT_CONTAINER_NAME = ${SERVICE_NAME}.canonical.com
 
-# golang is installed from the snap
-GO ?= /snap/bin/go
+GO ?= go
 
 # this repo contains external dependencies for an internal build
 VENDOR_BRANCH_URL ?= lp:~ubuntuone-pqm-team/serial-vault/+git/dependencies
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,12 @@ GOTAGS=-tags netgo
 
 SERVICE_NAME=serial-vault
 LOCAL_SERVICE_NAME = ${GOBIN}/${SERVICE_NAME}
-GO ?= go
+TAR_BASE_NAME = ${SERVICE_NAME}-v${VERSION}.tar
+TGZ_NAME = ${TMP}/${SERVICE_NAME}-payload-${VERSION}.tgz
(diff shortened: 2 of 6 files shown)


ols-vms.conf
```diff
@@ -0,0 +1,11 @@
+vm.architecture = amd64
+vm.release = focal
+
+[serial-vault]
+# PPAs for python-codetree and charm-tools
+apt.sources = ppa:mojo-maintainers/ppa, ppa:juju/stable
```
verterok, line 6: Since xenial there are no more deb packages for charm-tools in any PPA, this is now only available as snap.
Besides that, I don't see charm-tools being used by anything in serial-vault, you can probably drop the juju/stable ppa.

And a related issue: it's ok use focal? I see the current container in jenkaas is using xenial, will this break deployment on non-focal instances?

The current vms conf for the container is:
```
[serial-vault]
vm.class = lxd
vm.architecture = amd64
# Must be xenial because the thing is deployed on xenial and we live in the past.
vm.release = xenial
vm.update = True
vm.packages = git, python3-swiftclient, build-essential
jenkaas.secrets = swift/serial-vault:.config/swift/serial-vault
```
python3-swiftclient is required for the swift uploading bits, which is not required yet, but will not hurt to include it here, as it's a CI/Jenkins-only dependency.

ols-vms.conf
```diff
@@ -0,0 +1,11 @@
...
+vm.packages = @dependencies.txt, @dependencies-devel.txt, build-essential
```
verterok, line 7: build-essential is already included in dependencies-devel.txt, no need to include it here

Makefile
```diff
@@ -111,6 +115,21 @@ static-test:
 .PHONY: test
 test: unit-test static-test
 
+.PHONY: tarball
+tarball:
```
verterok, line 119: should this depend on some other target, like build? otherwise the assets will not be there...or I'm missing something?

verterok (commented):
looks good, some inline comments and a Q about the tarball target

Makefile
glower: yes, this makes sense, will fix this

ols-vms.conf
```diff
@@ -0,0 +1,11 @@
...
+
+# sideload snaps
```
verterok, line 9: Forgot about this one, no need to include the lines below if no snaps are required.

glower: I will use golang 1.13 from the snap because we don't have this version on xenial

ols-vms.conf
glower: Did some clean-up. I would like to upload directly from Jenkins, so added here

dependencies-devel.txt
```diff
@@ -1,3 +1,2 @@
 build-essential
 git
-golang-go
```
glower, line 3: golang is installed from the snap

.github/workflows/nodejs.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches: master
 jobs:
   build:
-    runs-on: ubuntu-latest
```
glower, line 8: `ubuntu-latest` is 18.04 in github actions, pinning it to focal

glower:
Added `publish-to-swift` script and the Makefile target from https://git.launchpad.net/turnip/

PR #306, opened 2020-09-17
Added additional configuration parameters for setting the port

glower:
Added support for portAdmin and portSign parameters to the configuration file. We need to have a possibility to override hardcoded ports for our deployment process

### DIFF
--- a/cmd/serial-vault/main.go
+++ b/cmd/serial-vault/main.go
@@ -59,13 +59,19 @@ func main() {
 	case "admin":
 		// Create the admin web service router
 		handler = service.AdminRouter()
-		address = ":8081"
+		address = datastore.Environ.Config.PortAdmin
+		if address == "" {
+			address = "8081"
+		}
 	default:
 		// Create the user web service router
 		handler = service.SigningRouter()
-		address = ":8080"
+		address = datastore.Environ.Config.PortSign
+		if address == "" {
+			address = "8080"
+		}
 	}
 
 	svlog.Infof("Starting service on port %s", address)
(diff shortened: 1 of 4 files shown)


settings.yaml.example
```diff
@@ -2,7 +2,12 @@ title: "Serial Vault"
 logo: "/static/images/logo-ubuntu-white.svg"
 
 # Service mode: signing or admin
-mode: signing
+mode: sign
+
+# Http ports for api and admin mode: defaults are 8081 for admin mode and 8080 for sign mode
+# You need to set both only if you want to run both services at the same time on the same machine
```
suligap, line 8: s/at the same time on the same machine/using the same settings.yaml file/ ?

settings.yaml.example
```diff
@@ -2,7 +2,12 @@ title: "Serial Vault"
...
+portAdmin: 8081
+portSign: 8080
```
suligap, line 10: Shouldn't we stick to using "signing" instead of sign? There's a `SigningRouter` for example.

cmd/serial-vault/main.go
```diff
@@ -59,13 +59,19 @@ func main() {
 	case "admin":
 		// Create the admin web service router
 		handler = service.AdminRouter()
-		address = ":8081"
+		address = datastore.Environ.Config.PortAdmin
```
suligap, line 62: Worth renaming it to `port` while we're at it?

suligap (approved):
LGTM with some nitpicks.

settings.yaml.example
glower: make sense, will rename it

PR #315, opened 2020-10-21
Added configuration to use rather staging or production store and SSO

glower:
Currently, the URLs for store API and SSO are hardcoded. Because of this, Serial-Vault on staging is connected to the production store and SSO login. To make SV more testable with the staging store we need to make these URLs configurable.


glower:
If the new config is not set, SV will use default URLs pointing to prod, so the believer should not change

PR #316, opened 2020-11-05
Added sentry reports

glower:
* Added additional configuration for the Sentry DSN
* Send sentry report on every 5xx service error
* Removed API keys from the report


cmd/serial-vault/main.go
```diff
@@ -27,10 +27,17 @@ import (
 	"github.com/CanonicalLtd/serial-vault/datastore"
 	"github.com/CanonicalLtd/serial-vault/service"
 	svlog "github.com/CanonicalLtd/serial-vault/service/log"
+	"github.com/CanonicalLtd/serial-vault/service/sentry"
 	logging "github.com/op/go-logging"
 )
 
-func init() {
```
glower, line 33: cant use init() here because it runs before main(), but we need config initialized at this point

config/config.go
```diff
@@ -95,5 +97,9 @@ func ReadConfig(settings *Settings, filePath string) error {
 		ServiceMode = settings.Mode
 	}
 
+	if settings.SentryDSN == "" {
+		settings.SentryDSN = os.Getenv("SENTRY_DSN")
```
glower, line 101: SENTRY_DSN is used on our infrastructure to set up sentry DSN, it's always set on the unit the service is deployed

service/middleware.go
```diff
@@ -83,3 +96,16 @@ var MiddlewareWithCSRF = func(inner http.Handler) http.Handler {
 
 	return CSRF(Middleware(inner))
 }
+
+func recoverWithSentry(hub *sentry.Hub, w http.ResponseWriter, r *http.Request) {
+	if err := recover(); err != nil {
+		log.Errorf("recover from error: %v", err)
+		w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+		w.WriteHeader(500)
```
suligap, line 104: Looks like no response body is actually sent out here. We might want to add an `InternalServerError` entry to `response.errors` and write it out here?

service/sentry/sentry.go
```diff
@@ -0,0 +1,82 @@
...
+
+func getFilteredEvent(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
+	for _, headerName := range headerFilter {
+		if _, ok := event.Request.Headers[headerName]; ok {
+			event.Request.Headers[headerName] = placeholder
+		}
+	}
+
+	if event.Request.QueryString != "" {
+		event.Request.QueryString = placeholder
```
suligap, line 66: It might be useful to retain the names of the params and filter out the values?

service/sentry/sentry.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	if event.Request.Data != "" {
+		event.Request.Data = placeholder
+	}
+
+	if event.Request.Cookies != "" {
+		event.Request.Cookies = placeholder
```
suligap, line 74: It might be useful to retain the names of the cookies and filter out the values?

suligap (changes requested):
Tested it and it's looking good. Except for the parts I commented on. Especially the missing response body when recovering from a panic I think.

service/sentry/sentry.go
```diff
@@ -0,0 +1,115 @@
...
+	if event.Request.QueryString != "" {
+		event.Request.QueryString = GetFilteredQueryString(event.Request.QueryString)
+	}
+
+	if event.Request.Data != "" {
+		event.Request.Data = placeholder
+	}
+
+	if event.Request.Cookies != "" {
+		event.Request.Cookies = placeholder
```
suligap, line 76: This is missing a `GetFilteredCookies()` call.

service/middleware.go
```diff
@@ -83,3 +97,24 @@ var MiddlewareWithCSRF = func(inner http.Handler) http.Handler {
...
+		log.Errorf("recover from error: %v", err)
+		debug.PrintStack()
+
+		w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+		e := response.ErrorInternal
+		w.WriteHeader(e.StatusCode)
+
+		// Encode the response as JSON
+		if err := json.NewEncoder(w).Encode(e); err != nil {
+			log.Printf("Error forming the signing response: %v\n", err)
```
suligap, line 112: This log message seems misleading. Maybe it should indicate that we're in the panic recovery middleware?

suligap (commented):
Two small things.

service/sentry/sentry.go
glower: right :) fixed

service/middleware.go
glower: fixed

suligap (approved):
Thank you!

PR #321, opened 2020-11-24
Added Github Actions for frontend tests and golang tests

glower:
Replace slow TravisCI with github actions for frontend and backend tests


glower:
With this change, we are using only Github Actions to run front-end tests, front-end build, and golang tests. This will make the CI process uniform and faster. We also disable TravisCI

Roadmaster (approved):
+1 let's merge this.

PR #328, opened 2020-11-24
Update ols-vms.conf

glower:


Roadmaster (approved):
+1 to bring in sync with the actually deployed units

PR #330, opened 2020-11-25
Move build process back to xenial

glower:


Roadmaster (changes requested):
I think we also need to update dependencies.txt to `s/postgresql-12/postgresql-9.5/` as 12 is not available on xenial. Optionally, this might be better: `s/postgresql-12/postgresql/`, postgresql is a metapackage which "always depends on the currently supported PostgreSQL database server version."

Roadmaster (approved):
thanks, +1 let's try this.

PR #334, opened 2020-11-26
Added filtering by username/email for the user page

glower:


Roadmaster (approved):
LGTM but I'm not great with javascript and React makes me want to die :)

PR #338, opened 2020-12-02
WIP: store sign model assertion in the database

glower:


glower:
In this patch, I added a new table similar to the table we have in SAS where we store assertions with signatures
Also added a migration script to create a signed assertion in the new database table for each model with an assertion
Also moved some code around

datastore/modelassertmanager.go
```diff
@@ -92,6 +127,134 @@ type ModelAssertion struct {
 	Modified      time.Time `json:"modified"`
 }
 
+// CreateModelAssertionHeaders returns the model assertion headers for a model
+func CreateModelAssertionHeaders(m Model) (map[string]interface{}, Keypair, error) {
```
pedronis, line 131: I understand that this wasmoved with the name from somewhere else, but in this package everything Create* operates mostly on the db. I wonder if this should just be called ModelAssertionHeadersForModel

datastore/modelassertmanager.go
```diff
@@ -130,6 +293,43 @@ func (db *DB) UpdateModelAssert(m ModelAssertion) error {
 	return nil
 }
 
+// UpsertSignedModelAssert creates or updates signed model assertion
+func (db *DB) UpsertSignedModelAssert(modelID int, assertion asserts.Assertion) error {
```
pedronis, line 297: we really need to make sure that we aren't storing an assertion with a lower revision than the current one, this is a central invariant of assertion behavior

service/model/actions.go
```diff
@@ -185,6 +186,34 @@ func assertionHeaders(w http.ResponseWriter, user datastore.User, apiCall bool,
...
+		keypair.KeyID,
+		keypair.SealedKey)
+	if err != nil {
+		log.Println(err)
+		response.FormatStandardResponse(false, "error-signing-assertions", "", err.Error(), w)
+		return
+	}
+
+	// store a signed model assertion in the database
+	err = datastore.Environ.DB.UpsertSignedModelAssert(model.ID, signedAssertion)
```
pedronis, line 210: if I understand correctly here we will just replace whatever we have in the db with a new signed assertion, for that to respect assertion rules we really need to check if have something already in the db and sign assertion with current revision + 1 as the newer replacement

service/pivot/handlers_pivot.go
```diff
@@ -87,7 +86,7 @@ func ModelAssertion(w http.ResponseWriter, r *http.Request) response.ErrorRespon
 	assertions := []asserts.Assertion{}
 
 	// Build the model assertion headers for the original model
-	assertionHeaders, keypair, err := assert.CreateModelAssertionHeaders(substore.FromModel)
+	assertionHeaders, keypair, err := datastore.CreateModelAssertionHeaders(substore.FromModel)
```
pedronis, line 89: if I undersand correctly this will continue to produce a new model for each pivoting device? in theory we should change this too to reuse the same model cached in the db if possible,  can be done later or not done given that pivot is a historical feature, nowadays we have remodel in principle. anyway wanting to make sure it's clear where we stand after this set of changes

pedronis (commented):
did a pass, didn't look at implementation details closely but made some high-level comments about behavior

datastore/modelassertmanager.go
glower: Yes, I moved this function from another package to avoid cyclic dependencies. It makes sense to rename this one

datastore/modelassertmanager.go
glower: Currently, we just override the entry in the database without checking the revision. In the UI user can increase or decrease the revision without any checks. 
One possibility to fix this is to remove the revision field from the UI and increment it in the backend before storing the assertion in the database.

service/model/actions.go
glower: Yes, this is correct. I can implement this check

service/pivot/handlers_pivot.go
glower: Yes, this is correct. I would change this in the next PR after we fixed all the problems with the model assertion

datastore/modelassertmanager.go
pedronis: yes, I think not letting the user set a revision seems correct.  It doesn't make sense from an assertion pov to create a different version (different values) at an older revision point

manage/manage.go
```diff
@@ -47,4 +49,10 @@ func openDatabase() {
 
 	// Open the connection to the database
 	datastore.OpenSysDatabase(datastore.Environ.Config.Driver, datastore.Environ.Config.DataSource)
+
+	// Opening the keypair manager to create the signing database
+	err := datastore.OpenKeyStore(datastore.Environ.Config)
```
glower, line 54: this change is needed for the migration code to get access to the key store for signing

service/model/actions.go
```diff
@@ -171,12 +172,16 @@ func assertionHeaders(w http.ResponseWriter, user datastore.User, apiCall bool,
...
 
 	// Check that the user has permissions to access the model
-	_, err = datastore.Environ.DB.GetAllowedModel(assert.ModelID, user)
+	model, err := datastore.Environ.DB.GetAllowedModel(assert.ModelID, user)
 	if err != nil {
 		log.Println(err)
 		response.FormatStandardResponse(false, "error-get-model", "", err.Error(), w)
 		return
 	}
+	// Increment the revision before it is stored in the database
```
pedronis, line 181: I don't understand this change, afaict from the callers assert is decoded from the JSON from the client(s), so this just increment by one what we got from the client instead of considering what we have in the db?

PR #343, opened 2021-01-11
Don't remove single quotes from headers

glower:
This is a revert for the change https://github.com/CanonicalLtd/serial-vault/pull/287
Wrong quoted serial-numbers in the header should be cleaned on the client-side

### DIFF
--- a/run-checks
+++ b/run-checks
@@ -133,7 +133,7 @@ if [ "$STATIC" = 1 ]; then
 
     echo Checking for ineffective assignments
     # ineffassign knows about ignoring vendor/ \o/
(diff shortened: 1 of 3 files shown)


run-checks
```diff
@@ -133,7 +133,7 @@ if [ "$STATIC" = 1 ]; then
 
     echo Checking for ineffective assignments
     # ineffassign knows about ignoring vendor/ \o/
-    ./bin/ineffassign .
```
glower, line 136: the usage of this tool changed, so it broke the build

pedronis (approved):
thank you

PR #391, opened 2022-09-26
modules: update libpq to support newer versions of postgres

IsaacJT:
Newer versions of Postgres (including that in 22.04) use scram-sha-256 authentication. Trying to connect to such a Postgres instance with the current version of libpq results in the following error message:

    2022-09-26 14:13:11Z CRITICAL serialvault "Error accessing the database: pq: unknown authentication response: 10"

Updating to the latest version of libpq fixes this.


alex-ramanau:
Hi! Thanks for your efforts, but lib/pq was updated in another [pr](https://github.com/canonical/serial-vault/pull/432). 
I think it should work now.

PR #430, opened 2024-06-19
Fix charm-publish target in charm/Makefile

verterok:
Skip git commit, tag and push if the tag already exists


bowenfan96 (approved):
+1, thank you for fixing this!

alex-ramanau:
Thanks verterok! Let's go ahead and merge it.

PR #435, opened 2024-08-08
Serial vault maintenance

alex-ramanau:
Update golang a to 1.22.5 + update prometheus_client/jwg-go


bowenfan96 (approved):
Thanks Alex for the updates! These are big version jumps and should be tested rigorously in staging. I think it's fine to land this and deploy there; I'm not aware of anything, including CI tests etc., that depends on SV staging.

PR #437, opened 2024-08-30
Fix setup and use dev version if tag is not in current ref/HEAD when building the tarball

verterok:


Makefile
```diff
@@ -6,7 +6,8 @@ LOCAL_BIN:=${TMP}/bin
 GOBIN=${BASEDIR}/bin
 
 GIT_REVISION = $(shell git rev-parse --short HEAD)
-VERSION ?= $(shell git describe --tags --abbrev=0)
+# get the exact tag if we are on "release mode" or append git-HASH if not yet on release
+VERSION ?= $(shell git describe --exact-match --tags 2>/dev/null || echo $(shell git describe --tags --abbrev=0)-git-$(shell git rev-parse --short HEAD))
```
maxiberta, line 10: Can't we just use `git describe --tags`'s output for non exact tag matches? E.g. `2.2.0-149-g395343d` ("149 commits after tag 2.2.0 on commit 395343d").

```sh
... || echo $(shell git describe --tags)
```

maxiberta (approved):
+1

Makefile
verterok: TIL! 
so, it was just a matter of removing `--abbrev=0`
thanks!

maxiberta (approved):
+1

PR #439, opened 2024-09-05
Include the md5 upload in the same shell execution as the tarball

verterok:
This uses the already present credentials in the environment required to upload to swift


maxiberta (approved):
+1

PR #442, opened 2025-01-09
Allow only substore models owned by the substore's brand

bowenfan96:


ruhajaved (approved):
Thanks for catching this!

alex-ramanau:
+1 from me! Thanks for fixing this, Bowen!

